PR #1, opened 2020-03-14
CBL-689: Rip out old implementation of OFFLINE

bmeike:
This commit removes all of the old platform logic for the OFFLINE state.  It also disables the NetworkReachabilityManager.  CBL-760 tracks re-enabling it

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -22,9 +22,7 @@ import android.support.annotation.GuardedBy;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
-import java.net.InetAddress;
 import java.net.URI;
-import java.net.UnknownHostException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -70,12 +66,9 @@
  * be notified of progress.
  */
 @SuppressWarnings({"PMD.GodClass", "PMD.CyclomaticComplexity", "PMD.TooManyMethods", "PMD.TooManyFields"})
-public abstract class AbstractReplicator extends NetworkReachabilityListener {
+public abstract class AbstractReplicator {
```
pasin: I think NetworkReachabilityListener is still needed. Basically the platform still needs to listener to the reachability change when the replicator's status of OFFLINE and inform the C4Replicator that the host becomes reachable via c4repl_setHostReachable() API.

common/test/java/com/couchbase/lite/ReplicatorOfflineTest.java
```diff
@@ -105,6 +106,7 @@ public void testChangeListenerEmptyArg() throws Exception {
     }
 
     @Test
+    @Ignore("Platform no longer has control of network reachablity")
```
pasin: Platform is still responsible for controlling network reachability. Platform needs to watch for the reachability when the replicator is OFFLINE and inform the C4Replicator to retry when the host is reachable.

bmeike: As noted in the description of this ticket, CBL-760

I though you all were complaining that my commits were to big.  Here's one divided into two, reasonable steps...

common/main/java/com/couchbase/lite/AbstractReplicator.java
pasin: I just saw CBL-760.

PR #2, opened 2020-03-25
Document Id

bmeike:
Abstract Replicator is thread safe
New C4 Database lifecycle support (except: CBL-787)
Document Id feature (still need tests: CBL-792)

This is still pretty big, I am afraid.  (I really am trying to make them smaller!).  Changing the lifecycle of the C4Replicator was a great opportunity finally, to make the Replicator thread-safe.
Note also that several of the complexity warning are gone!  Hurray!

There are still two outstanding problems: CBL-787 (resetCheckpoint doesn't work right) and CBL-792 (need automated tests for DocId)

All existing tests pass.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -15,7 +15,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-
 package com.couchbase.lite;
 
 import android.support.annotation.GuardedBy;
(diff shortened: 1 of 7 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -417,26 +432,21 @@ public boolean isDocumentPending(@NonNull String docId) throws CouchbaseLiteExce
...
-        }
-
         final Boolean pending;
-        try { pending = repl.isDocumentPending(docId); }
+        try { pending = lazyCreateC4Replicator().isDocumentPending(docId); }
         catch (LiteCoreException e) { throw CBLStatus.convertException(e, "Failed getting document pending status"); }
 
-        // ??? C4 call returns null if the native replicator is gone: Not yet implemented
-        return (pending != null) && pending;
+        if (pending == null) { throw new IllegalStateException("Pending doc ids is unexpectedly null"); }
```
pasin: What is the case that this could happen? It doesn't throw LiteCoreException when it happens?

pasin: Also the native one returns boolean not Boolean: 
https://github.com/couchbase/couchbase-lite-java-common/blob/64cd753e44fce6ef15bb3084cd940018c5c12032/common/main/java/com/couchbase/lite/internal/core/C4Replicator.java#L510

common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -233,54 +265,19 @@ public void documentEnded(
             final AbstractReplicator replicator = (AbstractReplicator) context;
             if (!replicator.isSameReplicator(repl)) { return; } // this handles repl == null
 
+            if (documents == null) {
```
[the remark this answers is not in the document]
bmeike: Yeah... I think that this cannot happen.  It is a callback from Core, though, and the callback is marked (perhaps incorrectly) @Nullable.  Unless I put this check in, PMD complains.  I figure it doesn't hurt.

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: Yes.  Nice catch.  This code is leftover from when the C4Replicator could be freed.  Fixing.

PR #13, opened 2020-10-28
Hotfix/2.8.1

bmeike:
There are two commits here:
1) This is the change that is driving 2.8.1: backward compatible default db directory.  This is the only code change in the product.
2) Fix the packaging for Java.  We had to do this by hand, in the 2.8.0 release.  This puts the support libraries into separate directories, by category

### DIFF
--- a/android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
+++ b/android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
@@ -59,8 +59,6 @@ public final class CouchbaseLiteInternal {
     private static final String LITECORE_JNI_LIBRARY = "LiteCoreJNI";
 
     private static final String TEMP_DIR_NAME = "CouchbaseLiteTemp";
-    private static final String DB_DIR_NAME = ".couchbase";
-
 
     private static final AtomicReference<SoftReference<Context>> CONTEXT = new AtomicReference<>();
     private static final AtomicReference<ExecutionService> EXECUTION_SERVICE = new AtomicReference<>();
@@ -143,7 +141,7 @@ public final class CouchbaseLiteInternal {
     @NonNull
     public static String makeDbPath(@Nullable String rootDir) {
         requireInit("Can't create DB path");
-        return verifyDir((rootDir != null) ? new File(rootDir) : new File(getContext().getFilesDir(), DB_DIR_NAME));
+        return verifyDir((rootDir != null) ? new File(rootDir) : getContext().getFilesDir());
     }
 
     @NonNull
--- a/common/main/java/com/couchbase/lite/AbstractDatabase.java
+++ b/common/main/java/com/couchbase/lite/AbstractDatabase.java
@@ -1734,17 +1734,17 @@ abstract class AbstractDatabase {
     // The fix is to use the original "real" default dir (the one used by all pre 2.8.0 code)
     // and to copy a database from the "2.8" default directory into the "real" default
     // directory as long as it won't overwrite anything that is already there.
-    private void fixHydrogenBug(@NonNull String rootDirPath, @NonNull String dbName) throws CouchbaseLiteException {
+    private void fixHydrogenBug(@Nullable String rootDirPath, @NonNull String dbName) throws CouchbaseLiteException {
         final String defaultDirPath = AbstractDatabaseConfiguration.getDbDirectory(null);
 
         // Both rootDir and defaultDir are canonical, so string comparison should work.
         // If this is not about a database located in the default directory, none of this is relevant.
-        if (defaultDirPath.equals(rootDirPath)) { return; }
+        if (!defaultDirPath.equals(AbstractDatabaseConfiguration.getDbDirectory(rootDirPath))) { return; }
         final File defaultDir = new File(defaultDirPath);
 
         // If this database doesn't exist in the 2.8 default dir, were'r done here.
         final File twoDotEightDefaultDir = new File(defaultDir, ".couchbase");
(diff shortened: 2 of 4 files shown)


android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
```diff
@@ -143,7 +141,7 @@ public static Context getContext() {
     @NonNull
     public static String makeDbPath(@Nullable String rootDir) {
         requireInit("Can't create DB path");
-        return verifyDir((rootDir != null) ? new File(rootDir) : new File(getContext().getFilesDir(), DB_DIR_NAME));
+        return verifyDir((rootDir != null) ? new File(rootDir) : getContext().getFilesDir());
```
pasin, line 144: The code before fixing and the code after fix could produce different db path. Is this backward compat with 2.8.0 (e.g. app that is used 2.8.0 without using 2.7.0 before)?

tools/fetch_litecore.sh
```diff
@@ -121,12 +121,19 @@ case "${OS}" in
...
+      mkdir "${SUPPORT_DIR}/libc++" 
+      mv -f lib/libgcc* "${SUPPORT_DIR}/libc++"
+      mv -f lib/libstdc* "${SUPPORT_DIR}/libc++"
+
+      mkdir "${SUPPORT_DIR}/libicu" 
+      mv -f lib/libicu* "${SUPPORT_DIR}/libicu"
+      rm -f "${SUPPORT_DIR}/libicu/"libicutest*
+
+      mkdir "${SUPPORT_DIR}/libz" 
+      mv -f lib/libz*.so* "${SUPPORT_DIR}/libz" || true # Only on CentOS6
```
pasin, line 136: I remember that there is one .a copied to the support directory as well. Will this fix exclude copying .a file?

android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
bmeike: It is *not* backwards compatible with 2.8.0.  It *is* compatible with 2.7.x.
If we have to make it compatible with 2.8.0 -- not an unreasonable suggestion -- we should keep the new directory an locate and copy existing dbs.  This seems like pretty big change to me.  Again, though, not unreasonable.  I think a better suggestion would be to get this out quickly and to advise people not to use 2.8.0.

tools/fetch_litecore.sh
bmeike: Good memory!  No.  The only .a is libz.a, so the pattern libz*.so* will not pick it up.

android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
pasin: Agreed. We should make sure to put this into the release note.

pasin (approved):

android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
borrrden: Is unlisting 2.8.0 an option?

borrrden (approved):

android/main/java/com/couchbase/lite/internal/CouchbaseLiteInternal.java
bmeike: I don't know.  That is a discussion we need to have.  I don't think there is a way to be 100% backwards compatible.  The most common case for 2.8.1 is going to be that there are two dbs with identical names one in the root directory, one is .couchbase.  Deciding which to keep probably requires direct human intervention.

common/main/java/com/couchbase/lite/AbstractDatabase.java
```diff
@@ -1726,4 +1728,35 @@ private void shutdownExecutors(
...
+        // If this database already exists in the real default directory,
+        // we can't risk trashing it. We just use the database in the real default
+        // directory and leave well enough alone.
+        // It is *always* possible to use 2.8 database, by specifying
+        // its directory explicitly.
+        if (exists(dbName, defaultDir)) { return; }
+
+        // This database is in the 2.8 default dir but not in the real
+        // default dir.  Copy it to where it belongs.
+        Database.copy(getDatabaseFile(twoDotEightDefaultDir, dbName), dbName, new DatabaseConfiguration());
```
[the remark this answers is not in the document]
pasin: Here is what I mean,
1. The partial copy will be rollbacked which mean delete.
2. The exception will be thrown.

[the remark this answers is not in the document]
bmeike: Ok.  If copying the database fails, I will delete the partial copy and throw an exception.
That means that any subsequent attempt to open the DB will also throw an exception... the application now has a poison pill.  Ok?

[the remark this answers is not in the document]
pasin: The upgrade / migration process always has some corner cases that I hope those will not happen.

pasin (approved):

borrrden (approved):
I'd like to see a test for this.  Looks like it can be simulated by creating things by specifying the "wrong" default directory and then viewing what happens when a database is created using no specified directory.

bmeike:
Yeah.  I'm with you.  Tests added.

bmeike:
Since this is urgent and since I've changed only test code, since approvals, I'm merging this

PR #14, opened 2020-12-02
CBL-1440: Fix FLSlice memory leak

bmeike:
This commit contains two changes:

1) FLSliceResult now implements AutoCloseable.  This is Java's contract that says "you must call close() on this, when you are done with it"
  Closing a core-owned object and failing to close a bindings-owned object just generate warnings. They no longer cause exceptions to be thrown.
2) The FLSliceResult heap block allocated in the JNI is now freed.  There are two cases:
   a) Java holds ownership of the FLSR: the allocated block is freed when Java frees is companion.
   b) Java passes ownership of the FLSR to Core: the allocated block is freed before the FLSR is returned to Core

### DIFF
--- a/common/main/cpp/native_c4prediction.cc
+++ b/common/main/cpp/native_c4prediction.cc
@@ -22,6 +22,8 @@
 using namespace litecore;
 using namespace litecore::jni;
 
+
+
 #ifdef COUCHBASE_ENTERPRISE
 
 #include <c4PredictiveQuery.h>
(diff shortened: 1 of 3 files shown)


borrrden (approved):
The switch of the boolean naming scared me at first but there is only one place that uses the `isCoreOwned` bool as true from what I can tell, so it is doing what it should.

pasin (approved):

PR #17, opened 2021-01-11
Candidate/2.8.x

bmeike:
Preliminary candidate for the 2.8.x release

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -96,7 +96,11 @@ public abstract class AbstractReplicator extends InternalReplicator {
         /**
          * The replication is actively transferring data.
          */
-        BUSY
+        BUSY,
+        /**
+         * Unrecognized replication state.
+         */
+        UNKNOWN
     }
 
 
@@ -292,8 +296,11 @@ public abstract class AbstractReplicator extends InternalReplicator {
     @NonNull
     private static ActivityLevel getActivityLevelFromC4(int c4ActivityLevel) {
         final ActivityLevel level = ACTIVITY_LEVEL_FROM_C4.get(c4ActivityLevel);
-        if (level == null) { throw new IllegalStateException("Unrecognized activity level: " + c4ActivityLevel); }
-        return level;
+        if (level != null) { return level; }
+
+        Log.w(LogDomain.REPLICATOR, "Unrecognized replicator activity level: " + c4ActivityLevel);
+
+        return ActivityLevel.UNKNOWN;
     }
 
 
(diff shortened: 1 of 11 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -96,7 +96,11 @@
         /**
          * The replication is actively transferring data.
          */
-        BUSY
+        BUSY,
+        /**
+         * Unrecognized replication state.
+         */
+        UNKNOWN
```
borrrden, line 103: Would it be prudent here to define this in such a way as it is not the "next status after busy" but rather something like -1 or 999?  That would defend against the case where for some reason another status is added in LiteCore that needs to map to the value after busy.

bmeike: It doesn't work like that, I don't think.  These things don't have values... they just are.  I could put it at the other end (before STOPPED).  But, FWIW, you can't do something like BUSY + 1....

bmeike:
All the tests have passed:
https://github.com/couchbaselabs/couchbase-lite-java-ee-root/pull/19

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: If I understand what you are saying, it doesn't work like that. These things don't have values... they are just typed values. I could put UNKNOWN at the other end (before STOPPED).

FWIW, though, you can't do something like BUSY + 1....

pasin: As a specific LiteCore version / commit will be used, what is the case to get UNKNOWN status? I think, if there is a new LiteCore status, we should map the status to the ActivityLevel (Public API) correctly.

borrrden: Oh, I see...

borrrden (approved):

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: I thought about this a lot, before I did it, @pasin .  You could convince me, though, to revert to throwing an exception.

I think, though, that it is quite reasonable to believe (purely abstractly, now) that there might be times at which the bindings cannot determine the replicators state.  It seem quite reasonable, to me, to report an unknown instead of making something up.

In our current implementation, users will never see the UNKNOWN state.  After some consideration, though, I think it is not unreasonable that they might.

borrrden (approved):

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: Fixed.  Per Pasin, any unrecognized Core state is now reported as "BUSY".

bmeike:
Added one more commit: changes to AbstractReplicatorConfiguration to make ReplicatorType visible from Kotlin.

pasin (approved):

common/main/java/com/couchbase/lite/AbstractReplicator.java
borrrden: This was more to prevent any accidents like with the stopping status, but BUSY vs UNKNOWN I don't much care.  LiteCore also defaults to BUSY now, so this is just an extra layer of protection.

PR #19, opened 2021-01-18
Only call setC4TmpDirPath once

bmeike:
The new version of c4_setTmpDir should (by my request!) only be called once.

This fix has the platform code calling it only at initialization.


pasin (approved):

borrrden:
Could you point out where `setC4TmpDirPath` is called now that it has been removed?

PR #20, opened 2021-01-19
Backport Heartbeat, MaxRetries and MaxRetryWaitTime

bmeike:
Backport Heartbeat, MaxRetries and MaxRetryWaitTime, now part of 2.8.4

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
@@ -32,17 +32,14 @@ import com.couchbase.lite.internal.utils.Preconditions;
  * Replicator configuration.
  */
 @SuppressWarnings({"PMD.GodClass", "PMD.TooManyFields"})
-abstract class AbstractReplicatorConfiguration {
+public abstract class AbstractReplicatorConfiguration {
 
     /**
      * Replicator type
      * PUSH_AND_PULL: Bidirectional; both push and pull
      * PUSH: Pushing changes to the target
      * PULL: Pulling changes from the target
-     *
-     * @deprecated Use AbstractReplicator.ReplicatorType
      */
-    @Deprecated
     public enum ReplicatorType { PUSH_AND_PULL, PUSH, PULL }
 
     //---------------------------------------------
@@ -52,7 +49,7 @@ abstract class AbstractReplicatorConfiguration {
     @NonNull
     private final Database database;
     @NonNull
-    private AbstractReplicator.ReplicatorType replicatorType;
+    private ReplicatorType replicatorType;
     private boolean continuous;
     @Nullable
     private Authenticator authenticator;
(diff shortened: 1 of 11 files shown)


borrrden (changes requested):
I feel badly for saying "you did too much" but to keep us aligned we should only be putting in heartbeat (and not the other two) since .NET and iOS are not scheduled to have it for 2.8.x.  The logic can stay, but is there a way to make it not visible externally?

bmeike:
Oh good grief.  Ya, sure.  Will fix.

common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -357,38 +360,43 @@ public final ReplicatorConfiguration setType(@NonNull AbstractReplicator.Replica
...
 
+    /**
+     * Return the max number of retry attempts made after connection failure.
+     */
+    final int getMaxRetries() {
+        return (maxRetries >= 0)
+            ? maxRetries
+            : ((continuous)
+                ? AbstractCBLWebSocket.DEFAULT_ONE_SHOT_MAX_RETRIES
+                : AbstractCBLWebSocket.DEFAULT_CONTINUOUS_MAX_RETRIES);
```
borrrden, line 383: @pasin Should these remain what they are in 2.8.x until 3.0?  From what I can tell these are the new defaults.

pasin: I think the changes will be done in 3.0, otherwise java/android will have different behavior from the other platforms.

bmeike: This function is not visible in 2.8.4 (x == 4, unless I am mistaken).  Whether these are the correct values seems a question only for 3.0.0, I think.

Unless you want me to actually remove them.

The amount of time this simple backport has taken is pretty staggering.  I wish we could find a more efficient way of doing this kind of stuff...

pasin: I might not understand this correctly. It seems like maxRetries is default to -1 so it means that getMaxRetries() will either use DEFAULT_ONE_SHOT_MAX_RETRIES or DEFAULT_CONTINUOUS_MAX_RETRIES?

pasin: DEFAULT_ONE_SHOT_MAX_RETRIES or DEFAULT_CONTINUOUS_MAX_RETRIES are for Lithium. It's OK to keep getMaxRetries() to use in Lithium but can you remove or comment out line 452 and 453 that use the value?

bmeike:
Removed all traces of maxretries and -interval

common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -144,7 +140,7 @@ public final ReplicatorConfiguration setChannels(@Nullable List<String> channels
     /**
      * Sets the the conflict resolver.
      *
-     * @param conflictResolver A conflict resolver.
+     * @param conflictResolver The replicator type.
```
pasin: Wrong description. Can you revert this back to A conflict resolver?

pasin (approved):
Have one comment but the rest looks good to me.

borrrden (approved):

common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
bmeike: Not a regression.  That comment has been there for ages.  WIll fix in 3.0

PR #22, opened 2021-02-04
Check heartbeat bounds

bmeike:
OkHttp explodes trying to set a ping interval > MAX_INT.  The explosion doesn't happen until the replicator actually tries to connect.

This code uses the same check that OkHttp uses, guaranteeing that if you can set the heartbeat in the config, it will work when the replicator connects.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
@@ -21,6 +21,9 @@ import android.support.annotation.Nullable;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.internal.Util;
 
 import com.couchbase.lite.internal.core.C4Replicator;
 import com.couchbase.lite.internal.core.CBLVersion;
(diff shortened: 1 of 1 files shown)


common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -21,6 +21,9 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.internal.Util;
```
pasin, line 26: It's internal!. I'm fine with this but just want to point it out. We probably needs to change a bunch of things when upgrading okhttp anyway incase the internal classes are gone.

pasin (approved):

common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
bmeike: Yes... It's a very, very fair point.

It was my opinion that it was less fragile to use the *exact* check that they use, than to try to duplicate a possibly moving target.  Absolutely open to opinions to the contrary

bmeike:
BTW, please note that, although not visible in the PR, this update includes the deprecation (simply marked "deprecated") of LogLevel.getValue()

borrrden (approved):

PR #25, opened 2021-06-22
v2.8.6: Android API 30 compatibility

bmeike:
The Android cert verification process calls, by reflection, a completely non-standard method.

I've added that method and refactored the standard method, so that both use the same code to either defer to the default, or to use the CBL checks.

Also, AsyncTasks have been deprecated in v30.  The Executor that powered them is no longer useful for CBL.  I've replaced it with a small, internal execution service.

Any other changes are just things to get the test code to run

### DIFF
--- a/android/debug/AndroidManifest.xml
+++ b/android/debug/AndroidManifest.xml
@@ -17,7 +17,7 @@ limitations under the License.
 -->
 <manifest
     xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.couchbase.lite.test"
+    package="com.couchbase.lite"
     >
 
     <uses-permission android:name="android.permission.INTERNET" />
--- a/android/main/java/com/couchbase/lite/internal/AndroidExecutionService.java
+++ b/android/main/java/com/couchbase/lite/internal/AndroidExecutionService.java
@@ -15,16 +15,15 @@
 //
 package com.couchbase.lite.internal;
 
-import android.os.AsyncTask;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.NonNull;
 
 import java.util.concurrent.Executor;
 import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.ThreadPoolExecutor;
 
 import com.couchbase.lite.LogDomain;
+import com.couchbase.lite.internal.exec.CBLExecutor;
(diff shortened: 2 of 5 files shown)


android/debug/AndroidManifest.xml
```diff
@@ -0,0 +1,34 @@
...
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
+
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+
+    <application
+        android:label="@string/app_name"
+        android:usesCleartextTraffic="true"
+        />
+</manifest>
```
pasin, line 34: There is also the same file in `android/androidTest` folder. What is the purpose of this one and the other one in `androidTest` folder?

android/main/java/com/couchbase/lite/internal/AndroidExecutionService.java
```diff
@@ -63,7 +62,7 @@ private CancellableTask(@NonNull Handler handler, @NonNull Runnable task) {
     // Constructor
     //---------------------------------------------
     public AndroidExecutionService() {
-        super((ThreadPoolExecutor) AsyncTask.THREAD_POOL_EXECUTOR);
+        super(new CBLExecutor());
```
pasin, line 65: Oh, it's good to know that the AsyncTask is not deprecated!

pasin (approved):
Look good to me. Have one question about the manifest file just for understanding.

android/debug/AndroidManifest.xml
bmeike: Yeah... that's mysterious.  Android builds the `debug` flavor of the library, for testing.  It then builds an entirely different application that contains the tests.  Both of those applications need to allow clear text traffic, in order for the tests (which use cleartext most fo the time) to pass.

android/main/java/com/couchbase/lite/internal/AndroidExecutionService.java
bmeike: It is deprecated!  It is not backed by a SynchronousQueue (a queue with max size 1!).  That makes it pretty unusable (as was the intention) for people like us.

PR #27, opened 2021-07-21
Fixes couchbase/couchbase-lite-java-ce-root#17

giallon:
add HAS_ATTACHMENTS flags in saveResolvedDocument.
Without it, the current revision lost attachment.

Fixes couchbase/couchbase-lite-java-ce-root#17


bmeike:
This was, very definitely a bug.  I believe it is this one: https://issues.couchbase.com/browse/CBL-2180
This, however, is not the fix.

Thanks very much for bringing the issue to our attention.

PR #28, opened 2021-09-02
CBL-2320: BlobOutputStream.read() is returning negative values.

bmeike:
FIx for CBL-2320

Must also be merged to java/release/lithium


LaurenNguyen14 (approved):
Based on our discussions on the issue, I think this should work

PR #32, opened 2021-09-24
Two bugs; one commit

bmeike:
This is fixes for:
CBL-2389 | Confirm LiteCore request for close, before connection is opened.
CBL-2395 | Kotlin FullTextIndexConfiguration?.create extension function does not support multiple expressions

### DIFF
--- a/common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
+++ b/common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
@@ -20,6 +20,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import java.util.Arrays;
+import java.util.List;
 import java.util.Locale;
 
 
@@ -30,6 +31,8 @@ public class FullTextIndexConfiguration extends IndexConfiguration {
     public FullTextIndexConfiguration(@NonNull String... expressions) {
         super(IndexType.FULL_TEXT, Arrays.asList(expressions));
     }
+
+    FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.FULL_TEXT, expressions); }
 
     /**
      * The language code which is an ISO-639 language such as "en", "fr", etc.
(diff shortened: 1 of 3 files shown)


common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
```diff
@@ -31,6 +32,8 @@ public FullTextIndexConfiguration(@NonNull String... expressions) {
         super(IndexType.FULL_TEXT, Arrays.asList(expressions));
     }
 
+    FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.VALUE, expressions); }
```
pasin, line 35: Should it be IndexType.FULL_TEXT?

pasin (changes requested):

common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
bmeike: OMG... OMG... What a catch!  Thanks!

bmeike: done

bmeike:
Please note: A complete fix for the root problem would, as noted in the related CBSE, require a timeout on silent connections.  Since that scenario is quite unlikely, and introducing a timer into sensitive code is a significant change, I am not going to do it.  The fix simply addresses the case in which the Core closes a connection that is not yet open.

pasin (approved):

bmeike:
Also note that it is quite difficult to write a dev test to confirm that this fix works: I have not added one.

I will have to set up a test network to try this out, probably using Jim's net-breaking proxy.  Will confirm results as soon as I have them.

It should be relatively easy for QE to test this scenario.  I hope that they will add such a test.

pasin:
I think OKHTTP might already have connection timeout.

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
```diff
@@ -175,7 +184,10 @@ public void onMessage(@NonNull WebSocket webSocket, @NonNull ByteString bytes) {
         public void onClosing(@NonNull WebSocket webSocket, int code, @NonNull String reason) {
             Log.d(TAG, "%s#OkHTTP closing: %s", AbstractCBLWebSocket.this, reason);
             synchronized (getPeerLock()) {
-                if (!state.setState(State.CLOSE_REQUESTED)) { return; }
+                if (!state.setState(State.CLOSE_REQUESTED))  {
+                    if (state.assertState(State.CLOSED)) { webSocket.cancel(); }
```
pasin, line 188: I couldn't think about the case that the state is already CLOSED while OKHTTP is closing. However it doesn't hurt to put the logic here for precaution.

PR #36, opened 2021-09-29
Live query java implementation

LaurenNguyen14:


common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
```diff
@@ -0,0 +1,40 @@
...
+    public C4QueryObserver newObserver(C4Query c4Query) {
+        // TODO: implement newObserver()
+        return null;
+    }
+
+    public void setEnabled(boolean enabled) {
+        //TODO: implement setEnabled()
+    }
+
+    public void getEnumerator(boolean forget) throws CouchbaseLiteException {
```
LaurenNguyen14, line 35: This should return the enumerator

bmeike (approved):
You are going to the a static constructor that hands the C interface implementation to a new instance of C4QueryObserver.  The constructor should be private.  You might not be able to make it private, but if you can’t you need to flag it as @VisibleForTesting

You are going to need a finalizer in there, in case close doesn’t get called.

I think the next step should be writing some tests.

Please go ahead and commit this much.

common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
```diff
@@ -0,0 +1,54 @@
...
+        long nCreate(long c4Query);
+        void nSetEnabled(long handle, boolean enabled);
+        void nFree(long handle);
+        long nGetEnumerator(long handle, boolean forget) throws LiteCoreException;
+    }
+
+
+    @VisibleForTesting
+    static NativeImpl nativeImpl = new NativeC4QueryObserver();
+
```
bmeike, line 23: I think you still need the static constructor...

bmeike (approved):
Still need the static constructor.  Looks really good, tho

PR #39, opened 2021-10-12
Live Query Implementation

LaurenNguyen14:
* remove LiveQuery, remove tests that use live query 
* register/de-register observer in AbstractQuery
* fix build fail


bmeike (changes requested):
There are several things here that need to be fixed.

PR #44, opened 2021-11-01
CBL-2435: Platform implementation of C4Query setParameters

LaurenNguyen14:


common/main/java/com/couchbase/lite/AbstractQuery.java
```diff
@@ -83,13 +83,23 @@
...
+     *
+     * @throws IllegalStateException    on failure to create the query (e.g., database closed)
+     * @throws IllegalArgumentException on failure to encode the parameters (e.g., parameter value not supported)
      */
     @Override
     public void setParameters(@Nullable Parameters parameters) {
-        synchronized (lock) {
-            this.parameters = (parameters == null) ? null : parameters.readonlyCopy();
+        synchronized (lock) { this.parameters = (parameters == null) ? null : parameters.readonlyCopy(); }
+        try {
```
bmeike, line 93: You don't need nested try blocks... you just need multiple catch blocks

Also, I think it would be better to wrap the smallest possible amount of code, in the try.

bmeike (approved):
Looks good.  Get rid of that double "try" and commit it.

PR #47, opened 2021-11-05
Release leaking certs

bmeike:
@pasin  found a couple of memory leaks.  This is the fix.

### DIFF
--- a/common/main/cpp/native_c4listener.cc
+++ b/common/main/cpp/native_c4listener.cc
@@ -574,6 +574,8 @@ JNICALL Java_com_couchbase_lite_internal_core_impl_NativeC4Listener_startTls(
         jboolean enableDeltaSync,
         jboolean requirePasswordAuth) {
     C4TLSConfig tlsConfig = {};
+    tlsConfig.rootClientCerts = nullptr;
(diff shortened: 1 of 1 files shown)


common/main/cpp/native_c4listener.cc
```diff
@@ -574,6 +574,8 @@ JNICALL Java_com_couchbase_lite_internal_core_impl_NativeC4Listener_startTls(
         jboolean enableDeltaSync,
         jboolean requirePasswordAuth) {
     C4TLSConfig tlsConfig = {};
+    tlsConfig.rootClientCerts = nullptr;
+    tlsConfig.certificate = nullptr;
```
borrrden, line 578: These two lines are redundant with the `{}` initializer (which basically memsets the whole thing to 0).

borrrden (approved):

common/main/cpp/native_c4listener.cc
bmeike: Ah... I wondered about that.  I wasn't sure that was guaranteed to be initialized.  Will remove.

LaurenNguyen14 (approved):

PR #51, opened 2021-11-19
Fix build_liteCore files 

LaurenNguyen14:
- Get rid of unreference code in build_litecore.sh
- Fix build_litecore.bat

### DIFF
--- a/tools/build_litecore.sh
+++ b/tools/build_litecore.sh
@@ -104,7 +104,7 @@ case $LIB in
    # works on centos6 and several version of OSX
    mbedcrypto)
       cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_POSITION_INDEPENDENT_CODE=1 ../../$MBEDTLS_DIR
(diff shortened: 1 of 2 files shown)


borrrden (approved):

tools/build_litecore.sh
```diff
@@ -104,7 +104,7 @@ case $LIB in
    # works on centos6 and several version of OSX
    mbedcrypto)
       cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_POSITION_INDEPENDENT_CODE=1 ../../$MBEDTLS_DIR
-      make -j $JOBS mbedx509 mbedcrypto mbedtls
+      make -j $JOBS mbedcrypto
```
bmeike, line 107: This is my bad.  We shouldn't change this, in Lithium.  It is not necessary and therefore, not minimal.  We *should* make the change in Helium, but not here.

Sorry about that.

bmeike (approved):
Please revert the one noted change that I mistakenly advised. 

Ok to commit, after that.

PR #52, opened 2021-11-23
CBL-2589: stop building mbedtls tests

bmeike:
Add flags to cmake build to prevent it from creating the mbedtls test suite.

### DIFF
--- a/tools/build_litecore.bat
+++ b/tools/build_litecore.bat
@@ -55,7 +55,7 @@ if "%lib%" == "LiteCore" (
 
 rem Works
 if "%lib%" == "mbedcrypto" (
-    "C:\Program Files\CMake\bin\cmake.exe" -G %VS_GEN% -DCMAKE_BUILD_TYPE=RelWithDebInfo ..\..\vendor\mbedtls || goto :error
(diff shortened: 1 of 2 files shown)


tools/build_litecore.bat
```diff
@@ -43,7 +43,7 @@ if not exist "%outputDir%" (
 
 rem Untested
 if "%lib%" == "LiteCore" (
-    "C:\Program Files\CMake\bin\cmake.exe" -G %VS_GEN% -DBUILD_ENTERPRISE=%entBuild% -DCMAKE_BUILD_TYPE=RelWithDebInfo ..\.. || goto :error
+    "C:\Program Files\CMake\bin\cmake.exe" -G %VS_GEN% -DENABLE_TESTING=OFF -DBUILD_ENTERPRISE=%entBuild% -DCMAKE_BUILD_TYPE=RelWithDebInfo ..\.. || goto :error
```
borrrden, line 46: I believe this is not strictly needed since I added this to be forced off in the LiteCore CMakeLists.txt but having it here hurts nothing.  Just an informational notice.  Ditto for the other non-mbedcrypto invocations.

borrrden (approved):

PR #58, opened 2022-01-28
Make sure there's no concatenation in logging

LaurenNguyen14:
Also, get rid of unnecessary wrappings. There will be another PR for the same fix in ee root.


bmeike (approved):
Looks great.  Thanks

PR #59, opened 2022-02-04
CBL-2417: Add cookies in header to cookie option

LaurenNguyen14:


common/main/java/com/couchbase/lite/internal/BaseImmutableReplicatorConfiguration.java
```diff
@@ -181,6 +182,22 @@ public void addConnectionOptions(@NonNull Map<String, Object> options) {
         final Map<String, Object> httpHeaders = new HashMap<>();
         // User-Agent:
         httpHeaders.put("User-Agent", CBLVersion.getUserAgent());
+
+        // Cookies:
+        final String cookieName = "Cookies";
+        if ((headers != null) && (!headers.isEmpty()) && headers.containsKey(cookieName)) {
```
LaurenNguyen14, line 188: I check for cookies separately to avoid deeply nested if statements

common/main/java/com/couchbase/lite/internal/BaseImmutableReplicatorConfiguration.java
```diff
@@ -181,6 +182,22 @@ public void addConnectionOptions(@NonNull Map<String, Object> options) {
...
+            //get current cookies value from option to add a new cookie
+            final String current = (String) options.get(C4Replicator.REPLICATOR_OPTION_COOKIES);
+            final StringBuffer cookieStr = (current == null) ? new StringBuffer() : new StringBuffer(current);
```
bmeike, line 191: The thing to do here is:
final Object existingCookies = options.get(C4Replicator.REPLICATOR_OPTION_COOKIES);
if (existingCookeis instanceof String) { ...
`null` is never an instance of anything, and doing it this way (an idiom) is type safe

common/main/java/com/couchbase/lite/internal/BaseImmutableReplicatorConfiguration.java
```diff
@@ -181,6 +182,22 @@ public void addConnectionOptions(@NonNull Map<String, Object> options) {
...
+
+            if (cookieStr.length() > 0) { cookieStr.append("; "); }
+            cookieStr.append(String.format(Locale.ENGLISH, "%s=%s", cookieName, headers.get(cookieName)));
```
bmeike, line 194: This seems to be going to a lot of work, in the case where the current cookies are null.  If that is the case, you don't need a string buffer and you don't need to append.  Just set the cookies!

common/main/java/com/couchbase/lite/internal/BaseImmutableReplicatorConfiguration.java
bmeike: Please check for `headers` being null, only once.
... and there is no reason to check to see if it is empty.

bmeike (changes requested):
Basically sound.  There's one idiom I'd like you to invorporate...

And I'd like you to tighten up your code, a lot.  This is pretty flabby.  Put it on a treadmill or something...

PR #60, opened 2022-02-18
CBL-2835: enable core logging domains

bmeike:
I'm not entirely clear on the goal for this ticket.  Would you please check to be sure I've done the right thing?


borrrden (commented):
@pasin What is the final word about the SQL domain?  That would be the only change here that I could think of if SQL is too spammy.

pasin:
I haven't looked it yet and told if it's too noisy or not. 

@jayahariv Can you share two logs at the info level and at the verbose level to see how noisy it is?

bmeike:
Not a single review.  Merging.

PR #63, opened 2022-04-06
CBL-1210: Update license header in source files

LaurenNguyen14:


bmeike (approved):
I count 320 files that are either .java or .kt.  If you can account for the missing 14, this is good to go

LaurenNguyen14:
> I count 320 files that are either .java or .kt. If you can account for the missing 14, this is good to go

The rest of the headers are added

PR #70, opened 2022-04-20
CBL-3023: correctly handle SocketTimeoutException

bmeike:
Correct handling of SocketTimeoutException.  Similar treatment of PortUnreachableException and NoRouteToHostException

@borrrden : will the use of domain `C4Constants.ErrorDomain.NETWORK` and code `C4Constants.NetworkError.HOST_UNREACHABLE` work in Hydrogen LiteCore?


borrrden:
No unfortunately, those constants are not present in Hydrogen.

borrrden (changes requested):
This won't work as is (thanks for the heads up).  Either Core or Java needs to change.  As it stands I am not sure what will happen if this is merged.  Probably turn into an "unknown error" of sorts.

borrrden:
Note:  We do have `kC4NetErrTimeout` (3) in the network domain in hydrogen, which is retryable.

LaurenNguyen14 (approved):

borrrden (approved):

PR #72, opened 2022-04-25
CBL-3044: Back Android Extended509TrustManager out of the Java code

bmeike:
The Android RTE uses an Extended trust manager that does not exist in generic Java.  Back out the code that uses it, for the Java release.


bmeike:
Note that, while this is a biggish change from the Android code, it is nearly identical to the Java 2.8.4 code.  There are two changes since 2.8.4:

1. Fix a bug that would give the certificate listener a list containing a single element, null
2. Add the extended pinned cert feature (check the entire chain, not just the leaf)

LaurenNguyen14 (approved):

borrrden (approved):

PR #75, opened 2022-05-06
CBL-3083: Make SSLExceptions recoverable errors

bmeike:
For CBSE-11861, make generic SSLExceptions into recoverable errors

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
+++ b/common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
@@ -43,7 +43,10 @@ import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLKeyException;
+import javax.net.ssl.SSLProtocolException;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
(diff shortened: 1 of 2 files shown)


LaurenNguyen14 (approved):

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
```diff
@@ -622,7 +625,7 @@ private void didClose(Throwable error) {
         }
 
         // SSLPeerUnverifiedException
-        if (error instanceof javax.net.ssl.SSLPeerUnverifiedException) {
+        if ((error instanceof SSLKeyException) || (error instanceof javax.net.ssl.SSLPeerUnverifiedException)) {
```
pasin, line 628: Comment above need an update.

pasin (approved):

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
```diff
@@ -638,6 +641,16 @@ private void didClose(Throwable error) {
             return;
         }
 
+        else if (error instanceof SSLProtocolException) {
+            closed(C4Constants.ErrorDomain.WEB_SOCKET, C4Socket.WS_STATUS_CLOSE_PROTOCOL_ERROR, null);
+            return;
+        }
+
+        if (error instanceof SSLException) {
```
borrrden, line 649: As a suggestion, I would put a log here if it doesn't exist somewhere else that indicates that this exception was transformed since the end result (user transient) is not logically related to the input exception.  This is probably good to do any time an exception is mapped, especially if the result is a code that is just used for its transientness.

borrrden (approved):

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
bmeike: Ok

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
bmeike: Discussed.  The logging of the exception at the top of this method is sufficient.

PR #78, opened 2022-05-12
Refactor C4Log to use proxy

LaurenNguyen14:


bmeike (approved):
Yeah!  Looks great!  Next one should go more quickly...

PR #82, opened 2022-05-17
CBL-3136: Failure during open may cause a crash

bmeike:
Catch any exception thrown during Core attempt to open a remote connection and clean up.


bmeike:
Please look at this carefully.  It is not exactly the same as the change I made in 3.1.0 and, since C4Socket in not yet unit-testable in 3.0, the change is pretty much impossible to test.

LaurenNguyen14 (approved):

borrrden (approved):
I looked at it carefully but all I see is an added try / catch that will call closed so it seems fine.

PR #83, opened 2022-05-20
CBL-3141: Create unit tests for collection

LaurenNguyen14:


common/test/java/com/couchbase/lite/BaseCollectionTest.java
```diff
@@ -0,0 +1,67 @@
...
+        testScope = Scope.getDefault(baseTestDb);
+        testCollection = testScope.getDefaultCollection();
+        Report.log(LogLevel.INFO, "Created base test Collection: " + testCollection);
+        assertNotNull(testCollection);
+    }
+
+    @After
+    public final void tearDownBaseCollectionTest() {
+        testScope.deleteCollection(testCollection);
+        testScope = null;
```
bmeike, line 31: Not useful...

common/test/java/com/couchbase/lite/CollectionTest.java
```diff
@@ -0,0 +1,283 @@
...
+        createSingleDocInCollectionWithId(docId);
+        verifyGetDocumentInCollection(docId);
+    }
+
+    //---------------------------------------------
+    //  Save Document
+    //---------------------------------------------
+
+    // base test method
+    private void testSaveNewDocInCollectionWithID(String docID) throws CouchbaseLiteException {
```
bmeike, line 55: private methods go last in the class...

... also, by conventions, only methods that are @Test have signatures that begin with "test".  Maybe use the suffix "Test" as you did in the base class?

bmeike (approved):
I didn't look at all the tests... I am assuming that they are, pretty much, cut and past copies of the ones in Database.
I love the way you did this... when we actually support collections, we just need to change how we get the testCollection and *all* of these tests will test real collections.  Way to go.

Couple small fixes, then push it!

PR #98, opened 2022-06-30
Revise fromJavaReplColls to use std::vector

bmeike:


common/main/cpp/native_c4replicator.cc
```diff
@@ -166,49 +166,49 @@ static bool pullFilterFunction(C4CollectionSpec, C4String, C4String, C4RevisionF
...
     if (!f_ReplColl_hasPullFilter)
-        return false;
+        return -1;
 
     jfieldID f_ReplColl_token = env->GetFieldID(cls_replColl, "token", "J");
     if (!f_ReplColl_token)
-        return false;
+        return -1;
 
+    int nColls = env->GetArrayLength(jColls);
```
pasin, line 211: Inside the for loop below, I understand that you would need to add a `C4ReplicationCollection` object to the vector using push_back(), unless the vector is initialized by size, otherwise the behavior could be undefined based on https://cplusplus.com/reference/vector/vector/operator[].

pasin (changes requested):

common/main/cpp/native_c4replicator.cc
bmeike: Wow.  Thanks!

common/main/cpp/native_c4replicator.cc
```diff
@@ -166,49 +166,50 @@ static bool pullFilterFunction(C4CollectionSpec, C4String, C4String, C4RevisionF
...
 
-// ??? Move the field stuff into initialization.
-static bool fromJavaReplColls(
+// ??? Move the field introspection into initialization.
+static int fromJavaReplColls(
         JNIEnv *env,
-        int nColls,
         jobjectArray jColls,
-        C4ReplicationCollection colls[],
+        std::vector<C4ReplicationCollection> colls,
```
borrrden, line 173: Without passing this by reference I fear this does not do what you think it does.  Passing by value actually makes a copy of the entire list, so you are just adding to a local one here, while leaving the original one unaffected.  I think what you want is to add it to the one you passed in right?  In that case use `std::vector<C4ReplicationCollection>& colls`

borrrden (changes requested):

bmeike:
Abandoning this request.  Many thanks to Jim and Pasin for fixing it.  The code will go in in another commit

PR #99, opened 2022-07-07
Fix test databases not deleted

jeffdgr8:
There are 4 databases that aren't cleaned up after being created in tests. After running tests repeatedly, these 4 databases will continue to duplicate. This properly handles deleting these databases, or in one instance, the base test DB fails to be deleted because a handle isn't closed.

Fixes https://github.com/couchbase/couchbase-lite-java-ce-root/issues/19.

Also fix a `Map` variable, writing to the wrong map and an incorrect error message.

### DIFF
--- a/common/main/java/com/couchbase/lite/Fleece.java
+++ b/common/main/java/com/couchbase/lite/Fleece.java
@@ -71,7 +71,7 @@ final class Fleece {
 
         throw new IllegalArgumentException(
             Log.formatStandardMessage(
-                "InvalidValueToBeDeserialized",
+                "InvalidCouchbaseObjType",
                 value.getClass().getSimpleName(),
                 SUPPORTED_TYPES));
     }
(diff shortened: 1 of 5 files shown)


common/main/java/com/couchbase/lite/Fleece.java
```diff
@@ -71,7 +71,7 @@ static Object toCBLObject(@Nullable Object value) {
 
         throw new IllegalArgumentException(
             Log.formatStandardMessage(
```
bmeike, line 73: I've accepted the rest of this... but this looks wrong to me.  What is your reasoning?

bmeike (commented):
Thanks a million for this.  Have added all the fixes except the one so commented.

bmeike:
Btw... FWIW, most of your previously requested changes made it in to release 3.0.2, which should go out very shortly.

PR #100, opened 2022-07-07
Fix/3.0.2

bmeike:
Fix several small bugs:
CBL-3361: Remove @ExperimentalCoroutines tags from KTX package
CBL-3360: Support null argument to COUNT()
CBL-3383: Fix IndexConfiguration API

### DIFF
--- a/common/main/java/com/couchbase/lite/FullTextIndex.java
+++ b/common/main/java/com/couchbase/lite/FullTextIndex.java
@@ -67,7 +67,7 @@ public class FullTextIndex extends Index {
     String getLanguage() { return language; }
 
     @Override
-    boolean isIgnoringDiacritics() { return ignoreDiacritics; }
+    boolean isIgnoringAccents() { return ignoreDiacritics; }
 
     @NonNull
     @Override
--- a/common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
+++ b/common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
@@ -34,7 +34,7 @@ public class FullTextIndexConfiguration extends IndexConfiguration {
         super(IndexType.FULL_TEXT, Arrays.asList(expressions));
     }
 
-    FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.FULL_TEXT, expressions); }
+    public FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.FULL_TEXT, expressions); }
 
     /**
      * The language code which is an ISO-639 language such as "en", "fr", etc.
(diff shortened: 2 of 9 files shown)


common/main/java/com/couchbase/lite/FullTextIndex.java
```diff
@@ -67,7 +67,7 @@ public FullTextIndex ignoreAccents(boolean ignoreAccents) {
     String getLanguage() { return language; }
 
     @Override
-    boolean isIgnoringDiacritics() { return ignoreDiacritics; }
+    boolean isIgnoringAccents() { return ignoreDiacritics; }
```
bmeike, line 70: @pasin : should this be public?   Is it part of the API?

common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
```diff
@@ -34,7 +34,7 @@ public FullTextIndexConfiguration(@NonNull String... expressions) {
         super(IndexType.FULL_TEXT, Arrays.asList(expressions));
     }
 
-    FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.FULL_TEXT, expressions); }
+    public FullTextIndexConfiguration(@NonNull List<String> expressions) { super(IndexType.FULL_TEXT, expressions); }
```
bmeike, line 37: @pasin : client code needs to be able to create one of these?

common/main/java/com/couchbase/lite/FullTextIndex.java
pasin: I crossed check with iOS and we haven't made them public.

common/main/java/com/couchbase/lite/FullTextIndexConfiguration.java
pasin: Yes

pasin (approved):

common/main/java/com/couchbase/lite/FullTextIndex.java
jeffdgr8: @bmeike @pasin I'm seeing this API as public in [ObjC](https://github.com/couchbase/couchbase-lite-ios/blob/master/Objective-C/CBLFullTextIndex.h#L30) and [Swift](https://github.com/couchbase/couchbase-lite-ios/blob/master/Swift/Index.swift#L81).

pasin: The API spec doesn't say that it's public but we could make it public as Objective-C is already public. I will make explicitly change to the API spec and inform all the parties.

PR #101, opened 2022-07-11
CBL-3341: Review and Complete Collection Tests

LaurenNguyen14:
- CBL-3357: Validate 32+ select items in the query fails or not
- Fix the error code for getting documents from a deleted collection

### DIFF
--- a/common/test/java/com/couchbase/lite/BaseCollectionTest.kt
+++ b/common/test/java/com/couchbase/lite/BaseCollectionTest.kt
@@ -22,9 +22,9 @@ import org.junit.Before
 import java.util.*
 
 open class BaseCollectionTest : BaseDbTest() {
-    protected var testCollection: Collection? = null
-    protected var testColName = ""
-    protected var testScopeName = ""
+    protected lateinit var testCollection: Collection
+    protected lateinit var testColName : String
+    protected lateinit var testScopeName : String
 
     protected val Scope.collectionCount
         get() = this.collections.size
--- a/common/test/java/com/couchbase/lite/BaseDbTest.java
+++ b/common/test/java/com/couchbase/lite/BaseDbTest.java
@@ -1663,13 +1663,13 @@ public abstract class BaseDbTest extends BaseTest {
         verifyBlob(doc.getBlob("doc-29"));
     }
 
-    protected Database openDatabase() throws CouchbaseLiteException { return verifyDb(createDb(getUniqueName("test_db"))); }
+    protected Database openDatabase() throws CouchbaseLiteException { return verifyOrDeleteDb(createDb(getUniqueName("test_db"))); }
 
     protected final void reopenBaseTestDb() throws CouchbaseLiteException { baseTestDb = reopenDb(baseTestDb); }
 
     protected final void recreateBastTestDb() throws CouchbaseLiteException { baseTestDb = recreateDb(baseTestDb); }
 
-    protected Database duplicateBaseTestDb() throws CouchbaseLiteException { return verifyDb(duplicateDb(baseTestDb)); }
+    protected Database duplicateBaseTestDb() throws CouchbaseLiteException { return verifyOrDeleteDb(duplicateDb(baseTestDb)); }
 
 
     // Some JSON encoding will promote a Float to a Double.
@@ -1678,7 +1678,7 @@ public abstract class BaseDbTest extends BaseTest {
         if (val instanceof Double) { return ((Double) val).floatValue(); }
         throw new IllegalArgumentException("expected a floating point value");
     }
-    private Database verifyDb(Database db) {
+    private Database verifyOrDeleteDb(Database db) {
         try {
             assertNotNull(db);
             assertTrue(new File(db.getPath()).getCanonicalPath().endsWith(C4Database.DB_EXTENSION));


common/main/java/com/couchbase/lite/AbstractDatabase.java
```diff
@@ -660,12 +660,22 @@ public long getCount() {
...
+     *
      */
+    @SuppressWarnings("PMD.PreserveStackTrace")
     @Deprecated
     @Nullable
     public Document getDocument(@NonNull String id) {
         try { return getDefaultCollectionOrThrow().getDocument(id); }
-        catch (CouchbaseLiteException e) { throw new IllegalStateException(Log.lookupStandardMessage("DBClosed"), e); }
+        catch (CouchbaseLiteException e) {
+            if (e.getCode() == CBLError.Code.NOT_OPEN) {
```
bmeike, line 672: Yeah.... nice.  You might have to check the Domain, too...

common/main/java/com/couchbase/lite/Collection.java
```diff
@@ -154,14 +154,21 @@ long getCount() throws CouchbaseLiteException {
...
     @Nullable
     public Document getDocument(@NonNull String id) throws CouchbaseLiteException {
         Preconditions.assertNotEmpty(id, "id");
         return withLockAndOpenDb(() -> {
             try { return Document.getDocument(this, id, false); }
-            catch (CouchbaseLiteException e) { Log.i(LogDomain.DATABASE, "Failed retrieving document: %s", id); }
-            return null;
+            catch (CouchbaseLiteException e) {
+                if (e.getCode() == CBLError.Code.NOT_FOUND) {
+                    Log.i(LogDomain.DATABASE, "Failed retrieving document: %s", id);
```
bmeike, line 167: I don't think you need to log this.  It will happen all the time...

... and, again, might want to check the domain

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
```diff
@@ -22,16 +22,18 @@ import org.junit.Before
...
+    protected var testScopeName = ""
 
     protected val Scope.collectionCount
         get() = this.collections.size
 
     @Before
     fun setUpBaseCollectionTest() {
-        testScope = baseTestDb.defaultScope
-        testCollection = testScope!!.getCollection(Collection.DEFAULT_NAME)
+        testColName = "test_collection"
```
bmeike, line 34: Consider using the name randomizer:  BaseTest.getUniqueName

common/test/java/com/couchbase/lite/BaseDbTest.java
```diff
@@ -55,6 +59,12 @@ public interface DocValidator extends Fn.ConsumerThrows<Document, CouchbaseLiteE
 
     public static C4Database getC4Db(@NonNull Database db) { return db.getOpenC4Database(); }
 
+    protected static <T extends Comparable<T>> void assertContents(List<T> l1, T... contents) {
+        List<T> l2 = Arrays.asList(contents);
```
bmeike, line 63: Huh.  I didn't know they had to be sorted to be considered equals.


This has the side effect of sorting the lists that are passed to it.  It would be better if it didn't do that... but at least it needs a comment.

common/test/java/com/couchbase/lite/BaseDbTest.java
```diff
@@ -1655,14 +1665,38 @@ protected final void verifyDocument(DictionaryInterface doc) {
         verifyBlob(doc.getBlob("doc-29"));
     }
 
+    protected Database openDatabase() throws CouchbaseLiteException { return verifyDb(createDb("test_db")); }
```
bmeike, line 1668: Please use the name randomizer, here.  We've had a a lot of trouble, in the past, with tests failing because of some cruft left around by previous tests.

common/test/java/com/couchbase/lite/BaseDbTest.java
```diff
@@ -1655,14 +1665,38 @@ protected final void verifyDocument(DictionaryInterface doc) {
...
+            assertNotNull(db);
+            assertTrue(new File(db.getPath()).getCanonicalPath().endsWith(C4Database.DB_EXTENSION));
+
+            return db;
+        }
+        catch (IOException e) {
+            deleteDb(db);
+            throw new AssertionError("Unable to get db path", e);
+        }
+        catch (AssertionError e) {
```
bmeike, line 1694: This shouldn't be a `catch` block, it should be a `finally` block

common/test/java/com/couchbase/lite/DbCollectionsTest.kt
```diff
@@ -165,248 +164,12 @@ class DbCollectionsTest : BaseCollectionTest() {
         assertEquals(0, scope.collectionCount)
     }
 
-    @Test(expected = IllegalStateException::class)
```
bmeike, line 168: hmmm... where did these tests go?

common/test/java/com/couchbase/lite/DbCollectionsTest.kt
```diff
@@ -146,7 +146,6 @@ class DbCollectionsTest : BaseCollectionTest() {
 
 // !!! TESTS BELOW NEED TO BE MOVED TO CollectionTest
 
-    @Ignore("CBL-3257: getScopeNames does not return default scope when it is empty")
```
bmeike, line 149: Oh!  Cool!  This is the one I just noticed, too.  Might be a merge conflict here.

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
```diff
@@ -64,5 +66,15 @@ open class BaseCollectionTest : BaseDbTest() {
         Assert.assertEquals(doc.id, savedDoc!!.id)
         return savedDoc
     }
+
+    @Throws(CouchbaseLiteException::class)
+    protected fun createDocsInBaseCollectionTest(n: Int) {
+        for (i in 0 until n) {
+            val doc = MutableDocument(String.format(Locale.US, "doc_%03d", i))
+            doc.setValue("key", i)
+            saveDocInBaseCollectionTest(doc)
```
bmeike, line 75: Is the "Base" collection the default collection?

common/test/java/com/couchbase/lite/CollectionTest.kt
```diff
@@ -0,0 +1,849 @@
...
+        verifyGetDocumentInCollection(docId)
+    }
+
+    // get doc from the same collection from a different database instance
+    @Test
+    fun testGetExistingDocWithIDFromDifferentDBInstance() {
+        val docID = "doc1"
+        val doc = MutableDocument(docID)
+
+        val coll1 = baseTestDb.createCollection("coll_1")
```
bmeike, line 55: I do worry that many of these tests will fail, if one of them does.  We've had that happen a lot in the past.  I really do suggest using the name randomizer.

common/test/java/com/couchbase/lite/CollectionTest.kt
```diff
@@ -0,0 +1,849 @@
...
+    fun testGetDocFromDeletedCollection() {
+        //store doc
+        createSingleDocInCollectionWithId("doc1")
+
+        //delete col
+        Report.log("test scope name $testScopeName")
+        baseTestDb.deleteCollection(testColName, testScopeName)
+
+        //should fail
+        testCollection!!.getDocument("doc1")
```
bmeike, line 83: run the formatter here?

common/test/java/com/couchbase/lite/CollectionTest.kt
```diff
@@ -0,0 +1,849 @@
...
+        val docID = "doc1"
+        val doc = createSingleDocInCollectionWithId(docID).toMutable()
+
+        val otherDB = duplicateDb(baseTestDb)
+        val collection =  otherDB.getCollection(testColName, testScopeName)
+        assertNotNull(collection)
+
+        try {
+            assertNotSame(collection, testCollection)
+            if (collection != null) {
```
bmeike, line 128: should you assert not null here?

common/test/java/com/couchbase/lite/CollectionTest.kt
```diff
@@ -0,0 +1,849 @@
...
+                assertEquals(1, collection.count)
+            }
+
+            // Update doc and store it into different instance
+            doc.setValue("key", 2)
+            TestUtils.assertThrowsCBL(CBLError.Domain.CBLITE, CBLError.Code.INVALID_PARAMETER) {
```
bmeike, line 134: sweet!

bmeike (commented):
This looks great.

I've left some comments but the most important one is that I'd like you to use the name randomizer whenever you create something that might conflict with the next test.  We had a lot of trouble, in the past, with one test failing and causing a cascade of other tests to fail, because the failing test left cruft around.

bmeike (changes requested):
Add the name randomizer and see what you think of the other comments.  Go ahead and commit.

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
LaurenNguyen14: No, it's the test collection that is created in BaseCollectionTest, but I'll change the name of this function to avoid confusion

common/test/java/com/couchbase/lite/BaseDbTest.java
LaurenNguyen14: it shouldn't be a finally block since we should only delete the database if there is an error in verifying the database

common/main/java/com/couchbase/lite/Collection.java
```diff
@@ -154,14 +154,20 @@ long getCount() throws CouchbaseLiteException {
...
      */
     @Nullable
     public Document getDocument(@NonNull String id) throws CouchbaseLiteException {
         Preconditions.assertNotEmpty(id, "id");
         return withLockAndOpenDb(() -> {
             try { return Document.getDocument(this, id, false); }
-            catch (CouchbaseLiteException e) { Log.i(LogDomain.DATABASE, "Failed retrieving document: %s", id); }
-            return null;
+            catch (CouchbaseLiteException e) {
+                if (e.getDomain().equals(CBLError.Domain.CBLITE) && e.getCode() == CBLError.Code.NOT_FOUND) {
```
bmeike, line 166: not requesting a change, here, but FYI:
CBLError.Domain.CBLITE will never be null.  That means that saying CBLError.Domain.CBLITE.equals(...) will never NPE.

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
```diff
@@ -22,16 +22,18 @@ import org.junit.Before
 import java.util.*
 
 open class BaseCollectionTest : BaseDbTest() {
-    private var testScope: Scope? = null
-    private var testCollection: Collection? = null
+    protected var testCollection: Collection? = null
+    protected var testColName = ""
+    protected var testScopeName = ""
```
bmeike, line 28: I don't know if you've come across "lateinit".  This might be one place that it was ok to use it.

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
bmeike: I see.  Cool!

common/test/java/com/couchbase/lite/BaseDbTest.java
```diff
@@ -55,6 +59,10 @@ public interface DocValidator extends Fn.ConsumerThrows<Document, CouchbaseLiteE
 
     public static C4Database getC4Db(@NonNull Database db) { return db.getOpenC4Database(); }
 
+    protected static <T extends Comparable<T>> void assertContents(List<T> l1, T... contents) {
```
bmeike, line 62: Oh!!  Very nice!

common/test/java/com/couchbase/lite/BaseDbTest.java
bmeike: Ahhh!!!! Gottcha.  This isn't a test, its a utility method.
In that case, I don't think it should delete the database at all.  It should return true or false and the db creator should be responsible for doing the right thing...

bmeike: Or.... if that would break a lot of legacy code, just rename it "verifyOrDeleteDb"...

bmeike (approved):

common/test/java/com/couchbase/lite/BaseCollectionTest.kt
bmeike: Cool!!!

bmeike (commented):
I haven't looked at all of the tests... that, of course, is the important part.  Please do walk through them and assure yourself that they are actually testing what they are meant to test.

... and I ask you to do this before, so if you've already done it, just commit.  It looks really good.

PR #103, opened 2022-07-17
Fix QueryTest.testWhereValued()

jeffdgr8:
I noticed the `QueryTest.testWhereNullOrMissing()` test case `TestCase(address.notNullOrMissing(), 2)` became `TestCase(address.isNotValued(), 2)` in `testWhereValued()` which should be incorrect, but the test was still passing. Correcting it as `TestCase(address.isValued(), 2)` passed as well. On further inspection, the `<` should actually be `<=`, which properly causes the current state to fail and the corrected to pass.


bmeike:
Fix in 3.1 release

PR #104, opened 2022-07-17
Fix QueryTest.testOrderBy()

jeffdgr8:
I noticed in `QueryTest.testOrdered()` that `array1` and `array2` being compared were both created from the same `firstNames` list. So the `assertArrayEquals()` check was actually comparing the original list to itself, rather than the `sorted` version.

After correctly using `sorted` as the expected array in the comparison, `testOrderBy()` was failing for a couple reasons. `verifyQuery()` was running the query twice as both `Enumerator` and `Iterable`, resulting in `firstNames` being 200 long, the same sorted list twice, one after the other (which doesn't match its sorted version with double names in a row).

After passing `false` for `runBoth` in `verifyQuery()`, I also found that the `Comparator`s were reversed for `ascending()` and `descending()`. After fixing this, the test is now passing with confidence!

### DIFF
--- a/common/test/java/com/couchbase/lite/QueryTest.java
+++ b/common/test/java/com/couchbase/lite/QueryTest.java
@@ -3305,8 +3305,8 @@ public class QueryTest extends BaseQueryTest {
 
         List<String> sorted = new ArrayList<>(firstNames);
         Collections.sort(sorted, cmp);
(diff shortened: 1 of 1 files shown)


bmeike:
FIx in 3.1 release

PR #106, opened 2022-08-02
Improved FLSliceResult management

bmeike:
Refactor FLSliceResult to avoid the crazy mallocing


bmeike:
This change affects a whole bunch of files.  I don't think it is anything like necessary to read them all.  Probably the C-language stuff is most important.

borrrden (approved):
I understand that now you've decomposed slices into their respective parts when crossing the JNI boundary, in order to avoid the weird malloc, and introduced a Java class to hold them on the managed side.  Seems fine!

LaurenNguyen14 (approved):

PR #111, opened 2022-08-15
8.5.2 Test use collection API on collection deleted in a different database instance...

LaurenNguyen14:
.. and fix the issue that’s causing several filter tests with docIds and channels to fail (ignored tests tagged with this ticket [CBL-3424](https://issues.couchbase.com/browse/CBL-3424)) because option was not handled correctly in native_c4Replicator. The solution is to retain collection options until c4replicator is created.

### DIFF
--- a/common/test/java/com/couchbase/lite/CollectionListenerTest.kt
+++ b/common/test/java/com/couchbase/lite/CollectionListenerTest.kt
@@ -18,7 +18,9 @@ package com.couchbase.lite
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNotEquals
+import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
+import org.junit.Ignore
 import org.junit.Test
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.Executor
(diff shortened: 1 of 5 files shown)


pasin (commented):
* Suggest to have some detail about the PR in the PR's description. The description for a fix of an issue should include the info about problem and the solution so that the reviewers can have some context when reviewing the PR.

* For all of the Ignores, if possible, tagged them with CBL tickets?

common/test/java/com/couchbase/lite/CollectionListenerTest.kt
```diff
@@ -53,6 +55,114 @@ class CollectionListenerTest : BaseCollectionTest() {
...
+    @Ignore("Native crash")
+    @Test
+    fun testAddChangeListenerToDeletedCollection() {
+        baseTestDb.deleteCollection(testColName, testScopeName)
+        testCollection.addChangeListener(testSerialExecutor) {}
+    }
+
+
+    // Test that addChangeListener to a collection deleted from a different db instance get a warning message and doesn't throw exception
+    @Ignore("Native crash")
```
bmeike, line 68: The problem with this is that it isn't actionable: there's no real way to tell when the @Ignore can be removed.  I think the comment on the Ignore should be a ticket number....

common/test/java/com/couchbase/lite/CollectionTest.kt
```diff
@@ -66,6 +66,20 @@ class CollectionTest : BaseCollectionTest() {
         }
     }
 
+    // Test get doc from collection that is deleted in a different database instance
```
bmeike, line 69: Again... It would be so much more helpful if the comment gave the expected result: "verify that getting a document that has been deleted from a different database causes a CBL exception"

bmeike (approved):
Looks pretty good to me.  Very thorough testing.
As mentioned, though, I would like specific ticket ids in the Ignores and if you can cite the expected results in the test comments, that would be awesome.

pasin (approved):

PR #115, opened 2022-08-29
Fix query tests according to specs

LaurenNguyen14:
Test for CBL-3135 (core ticket: CBL-3626)


LaurenNguyen14:
After the fix in testStringQueryJoinWithCollections(), the test no longer throws exception but the query result is wrong. The test is tagged with the corresponding ticket.

bmeike (approved):
Looks good, with the exception of the test of the JSON on line 3400.  I don't think there is any guarantee about the order in which things will get put into the string...

PR #117, opened 2022-09-09
CBL-3666: Connectivity Manager can be registered multiple times.

bmeike:
Hotfix for CBSE-12499

When a continuous replication goes offline, it registers an observer to watch for changes in connectivity. The code used to allow a single observer to be registered multiple times. If this happened often enough, the system would throw an uncheck exception: TooManyRequestsException, which will cause the thread to fail.

The heart of the fix is line 313, checking to see if the observer is already in the map.  The try-catch blocks are just bullet-proofing.

### DIFF
--- a/android/main/java/com/couchbase/lite/internal/AndroidConnectivityManager.java
+++ b/android/main/java/com/couchbase/lite/internal/AndroidConnectivityManager.java
@@ -114,10 +114,15 @@ public class AndroidConnectivityManager implements NetworkConnectivityManager {
 
         @Override
         public void start() {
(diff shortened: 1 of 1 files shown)


borrrden (approved):
I'm assuming this is a simple merge of something that already went out

bmeike:
@borrrden : it is not.  I put it in Helium first.  It seems to work but, as noted, I do not have a good way to test it very thoroughly.  I assumed that we would push the relevant customer to update to 3.0, so I put it in 3.0, next.  That has not been released.  I just got notification from Dan that we'll need a 2.8 VF.  This will produce that VF.

PR #118, opened 2022-09-11
CBL-3666: Catch two more possible exceptions

bmeike:
These two try-catch blocks were, erroneously, left out of the previous commit.


bmeike:
I am, temporarily, overriding the review requirement for this commit, in order to get the build started.  If I do not get approval, I will roll back the commit and will not release the artifacts

PR #121, opened 2022-10-21
Dev

bmeike:
Use new Defaults
Add newly specified Index creation signature to Collection
Improved logging of active processes
Replicator should default to one-shot
Better document flag handling
Clean up document change handler per 3.0.5


common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -133,6 +123,9 @@ private static Map<Collection, CollectionConfiguration> copyConfigs(
     @Nullable
     private Database database;
 
+    // Yuggg: what a kludge.
+    boolean maxAttemptsSet;
```
borrrden, line 127: You could mimic the C# Nullable<bool> type I suppose which would at least keep this kludgery contained inside a separate class.

borrrden (approved):

PR #126, opened 2022-11-02
Begin moving tests off deprecated APIs

bmeike:
This is the MASSIVE collection of changes that begin to move tests off the deprecated APIs.
None of these changes affect any production code.

I've tried a couple times to make this commit smaller and was pretty unsuccessful.  Subsequent commits, however, really should be smaller.


borrrden (approved):
If the tests pass that's enough for me.  I went through several hundred lines of what basically just looked like renaming, changing whitespace, and refactoring.  The goal here is understood though.

pasin (approved):

PR #127, opened 2022-11-04
Make ListenerToken abstract

bmeike:
A couple of small tweaks to production code, suggested by work on tests.

Notably, ListenerToken is now an abstract class

These are all of the changes that affect production code


common/main/java/com/couchbase/lite/ListenerToken.java
```diff
@@ -30,7 +30,7 @@
 /**
  * Base class for a removable subscription to an observable.
  */
-public class ListenerToken extends AtomicBoolean implements AutoCloseable {
+public abstract class ListenerToken extends AtomicBoolean implements AutoCloseable {
```
pasin, line 33: This is not a big deal but basically users can use to store boolean.

pasin (approved):

PR #131, opened 2022-11-12
Remove use of deprecated APIs from most CE tests

bmeike:
This commit contains no changes to production code: It is all tests.


borrrden (approved):
This PR also seems to include changes from the other small production one, contrary to the description.  It is understandable why that would happen though.

PR #137, opened 2022-11-21
CBL-3844: Add documentation that clarifies the behavior of AutoPurge

bmeike:
Also, accommodate bizarrely slow Windows Java


bmeike:
Had to check this, myself: The kotlin setters for the auto-purge property are in the CE and EE repos, respectively, because they are different, depending on the version.

borrrden (approved):

PR #139, opened 2022-11-23
Add tests for legacy replicator configuration factory

bmeike:


common/test/kotlin/com/couchbase/lite/DeprecatedConfigFactoryTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        assertEquals(config1.database, config2.database)
+        assertEquals(setOf(testCollection.database.defaultCollection), config2.collections)
+    }
+
+    // Create from a source with default collection, explicitly specifying a non-default collection
+    @Test
+    fun testReplicatorConfigFromCollectionWithDefaultAndOther() {
+        val config1 = ReplicatorConfigurationFactory
+            .newConfig(testEndpoint, mapOf(testCollection to CollectionConfiguration()))
+        val config2 = config1.create()
```
pasin, line 91: Will config2 has the same collection config (which has testCollection) as the config1?

common/test/kotlin/com/couchbase/lite/DeprecatedConfigFactoryTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // Create from a source without default collection,
+    // explicitly specifying a non-default collection should fail
+    @Test(expected = IllegalArgumentException::class)
+    fun testReplicatorConfigFromCollectionWithoutDefault() {
+        testCollection.database.defaultCollection!!.delete()
+        val config1 = ReplicatorConfigurationFactory
+            .newConfig(testEndpoint, mapOf(testCollection to CollectionConfiguration()))
+        config1.create()
```
pasin, line 105: I think I'm confused here. Why an exception is thrown here when there is no default collection?

common/test/kotlin/com/couchbase/lite/DeprecatedConfigFactoryTest.kt
bmeike: Not quite.  ReplicatorConfiguration.create is deprecated.  It cannot handle arbitrary collections.  If the source replicator configuration contains collections other than the default collection, it does not know how to copy them.

This situation is easily avoided by not mixing ReplicatorConfiguration.create with ReplicatorConfiguration.newConfig (the latter being the new, collection enabled, signature)

common/test/kotlin/com/couchbase/lite/DeprecatedConfigFactoryTest.kt
bmeike: In this case, the database in config1 does not have a default collection.  It is not possible to configure, e.g., a document ID filter using ReplicatorConfiguration.create, because that attempts to configure the document ID filter on the default collection.  Unlike the situation above, where the default collection could be configured, here it cannot be configured.  It must be an error.

pasin (approved):

bmeike:
I've improved the first test to which @pasin refers to show that the default configuration *IS* configured, and improved the comments on the second.

PR #140, opened 2022-11-23
Package protect FleeceValue.withContent

bmeike:
This trivial change makes FLValue.withContent package protected.  It was public so that it could be used from tests.
Changing to this implementation removes a Lint warning and returns the method to package visibility.


bmeike:
Please review https://github.com/couchbaselabs/couchbase-lite-java-ee-root/pull/77 first...

borrrden (approved):

PR #144, opened 2022-12-01
CBL-1727: remove deprecated ReplicatorConfiguration.ReplicatorType

bmeike:
The type was deprecated in lithium.  It can go now.


pasin (commented):
This is a breaking change, and the helium version will be 3.1 which is not a major release. I think we cannot remove the deprecated methods?

bmeike:
I guess that it right, huh.  Ok.  Reverting.

PR #150, opened 2022-12-13
Pr/cbl 3941

bmeike:
I verify (after extensive discussion with @pasin ) that Java is doing the right thing.  It passes the *entire* value of the Set-Cookie header to Lite Core.

In any subsequent request in which the socket is in the UNOPENED or OPENING state (that is, not OPEN, CLOSING or CLOSED), it will add any cookies that it gets from LiteCore in the `options` parameter passed in the socket creation callback, as well as any cookies it gets from Lite Core, in a call to `c4db_getCookies`.

In both of these two cases the Java code expects that the string that it gets from Lite Core will *NOT* be cookies as they appear in a request header but will, instead, be <cookie-name>=<cookie-value>(;<cookie-name>=<cookie-value>)*

As long as Lite Core is correctly parsing and expiring cookies, the Java code is doing the right thing.

En-passant, I've eliminated a bunch of unnecessary locking and moved a test that should be common, out of the EE only tests.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractDatabase.java
+++ b/common/main/java/com/couchbase/lite/AbstractDatabase.java
@@ -1233,13 +1233,22 @@ abstract class AbstractDatabase extends BaseDatabase
 
     // - Cookie Store:
 
-    void setCookie(@NonNull URI uri, @NonNull String setCookieHeader) {
+    // We send the entire Set-Cookie string to Lite Core: e.g.,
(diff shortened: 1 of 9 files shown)


bmeike:
The specific issue at hand is GCLB Cookies.  I suspect that I will have to include them in every request, not just when the socket is being opened?

bmeike:
Pasin points out that we only have one Request per session: All good.

borrrden (approved):

PR #157, opened 2023-01-25
CBL-4121: Accept parent domain cookies

bmeike:
... also fix a JNI memory leak in Listener.start()

Who knew that such a small thing would require touching 20 different files...

### DIFF
--- a/common/main/cpp/native_c4listener.cc
+++ b/common/main/cpp/native_c4listener.cc
@@ -394,6 +394,8 @@ static C4Listener *startListener(
     config.directory = path;
     config.allowCreateDBs = allowCreateDBs;
     config.allowDeleteDBs = allowDeleteDBs;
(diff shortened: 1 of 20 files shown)


common/main/cpp/native_c4listener.cc
```diff
@@ -394,6 +394,8 @@ static C4Listener *startListener(
     config.directory = path;
     config.allowCreateDBs = allowCreateDBs;
     config.allowDeleteDBs = allowDeleteDBs;
+    config.allowCreateCollections = false;
+    config.allowDeleteCollections = false;
```
borrrden, line 398: Since this was just a smoke test, it is not needed for the final code.  These two will already be false by virtue of you creating the struct via `{}`.

pasin (approved):

borrrden (approved):

PR #159, opened 2023-02-17
CBL-4215: Fix broken Kotlin tests

bmeike:


common/main/java/com/couchbase/lite/Collection.java
```diff
@@ -136,16 +136,19 @@ static Collection getDefaultCollection(@NonNull Database db) throws CouchbaseLit
...
     /**
-     * Get scope
+     * Return the collection name
      */
     @NonNull
-    public Scope getScope() { return new Scope(c4Collection.getScope(), db); }
+    public String getName() { return c4Collection.getName(); }
+
+    @NonNull
+    public String getFullName() { return c4Collection.getScope() + "." + c4Collection.getName(); }
```
pasin, line 151: Can it be non-public?

pasin (changes requested):

common/main/java/com/couchbase/lite/Collection.java
bmeike: Yes... but why?  You are about to request that it be public... That seems just so mean to our clients....

pasin: I knew I want to have it public for all but not for one. Until then, I suggested that it is non-public.

pasin (approved):

common/main/java/com/couchbase/lite/Collection.java
bmeike: Done.  You are just a meany!!

pasin: Sorry for that. I need to keep the rule that we all jump at the same time if possible.

PR #164, opened 2023-03-08
CBL-4283: Make Collection.collectionChangeFlow's executor argument optional

bmeike:
Collection.collectionChangeFlow should default its argument, the Executor, to null, making it optional

Ticket: https://issues.couchbase.com/browse/CBL-4283
Targeting the RC2 release
This is very low risk.


borrrden (approved):
I'm assuming this is not an existing API that is public (rather a new one coming in) because otherwise this is breaking.

bmeike:
I don't think it is breaking....
These used to work: `myCollection.collectionChangeFlow(myExecutor)`: `myCollection.collectionChangeFlow(null)`
Both of those still work.  In addition, this works now: `myCollection.collectionChangeFlow()`

PR #174, opened 2023-04-13
CBL-4423: Stop LiveQueries before closing a DB

bmeike:
This commit is a little bigger than it really ought to be:

- In addition to adding a test for the bug fix, I changed several of the other tests to avoid a compiler warning.  Instead of try-finally, they are now try-with-resources.
- There is a publicly visible comment on AbstractQuery.explain that was just really broken
- there were two locks missing in the call to AbstractQuery.addChangeListener: one to protect getC4QueryLocked, and one to protect C4QueryObserver.setEnabled

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
+++ b/common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
@@ -137,6 +137,8 @@ public class C4QueryObserver extends C4NativePeer {
         if ((results != null) || (err != null)) { callback.onQueryChanged(results, err); }
     }
 
+    // ??? This is called only from LiteCore.
+    // Should it still be guarded by the db lock?
     @Nullable
     private C4QueryEnumerator getEnumerator() throws LiteCoreException {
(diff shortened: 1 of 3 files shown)


common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
```diff
@@ -137,6 +137,8 @@ void queryChanged() {
         if ((results != null) || (err != null)) { callback.onQueryChanged(results, err); }
     }
 
+    // ??? This is called only from LiteCore.
+    // Should it still be guarded by the db lock?
```
pasin, line 141: I checked iOS code and iOS doesn't lock the database when calling `c4queryobs_getEnumerator`.

pasin (approved):

common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
bmeike: Jens says it need to be done...

> These functions follow the general rule: only one thread at a time can call any of these functions with a specific C4Database reference.
> ...
> Queries
>    ...
>     c4queryobs_* (except c4queryobs_create which is thread-safe)

PR #176, opened 2023-04-19
Temporary workaround for CBL-4435: remove stopped replicator from DB active process list

bmeike:
This is just so I can get the dev tests to work.  I don't think it is a full fix for the issue.

Also: fix unsafe call to getOpenC4CollectionLocked() and replace C4ReplicatorStatus.copy with a more idiomatic copy constructor.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -382,6 +382,15 @@ public abstract class AbstractReplicator extends BaseReplicator
         }
     }
 
+    // When a replicator is closed it is detached from its from LiteCore and receives no further status updates.
+    // If it is not already in the STOPPED state, it will never be in the stopped state.  That means that
+    // a database that holds a reference to it can never close.
+    // Let's just tell the db forget about it.
+    public void close() {
+        getDatabase().removeActiveReplicator(this);
+        super.close();
+    }
+
     @NonNull
     @Override
     public String toString() {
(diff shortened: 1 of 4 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -382,6 +382,15 @@ public void removeChangeListener(@NonNull ListenerToken token) {
         }
     }
 
+    // When a replicator is closed it is detached from its from LiteCore and receives no further status updates.
```
pasin, line 385: As close() is the public API method, mention about LiteCore could be confusing to users. Maybe move the internal comment inside the method itself?

Also calling close() while the replicator is not stopped, seems to be dangerous and could cause unexpectedly crash I guessed. I think it might be worth mentioning about that as well.

pasin (approved):

borrrden (approved):
Other than Pasin's comment

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: Good points:
- That comment is not a javadoc comment.  It is visible only to people who are reading the code...  like you and me ;-P
- This is not the right solution.  I have a ticket to fix it for real.  None the less, I think it would be hard to cause a crash....  Maybe if you close while a push filter is running...

The final solution may be to disallow closing *unless* the replicator is stopped...

PR #180, opened 2023-05-03
CBL-3889: Get URLEndpointListener tests to the new API

bmeike:
Also a fix to the LiveQueries class that makes it more sensible.  This is the only change that affects production.


pasin (approved):
I couldn't find changes that update URLEndpointListener tests to the new API. Maybe the PR title is not accurate?

PR #182, opened 2023-05-19
CBL-4409: Remove error-prone 'expected' annotation

bmeike:
Also: CBL-4344: Close autocloseables

This really should be two separate commits.  The juggling necessary to do that, though would just take us all forever.  Please pardon my combining them


bmeike:
btw, the problem with the "@Test(expected=...) annotation, which will make a test fail *unless* it throws the expected exception, is that the exception can come from anywhere in the test, not only the intended location.

Also, "expected" cannot check a CBL exception to be sure that it has the expected domain and code

pasin (approved):

PR #184, opened 2023-05-25
CBL-4451: Support for Apple Silicon

bmeike:
This is the first step towards apple silicon support.

The plan (for MacOS) is that there is a single, universal LiteCore lib, but two JNI libs, one for x86_64, the other for aarch64.  This is because I despair of getting gradle to run lipo
    
  ```
                            .
                            |
                          macos
               /            |             \
           aarch64        x86_64         universal
             |              |                |
       LiteCoreJNI      LiteCoreJNI       LiteCore

```  
    
    There are 3 changes necessary:
    - The plugin for Jim's Core puller now puts the litecore binary for MacOS into a directory called "universal"
    - The build system processes the whole OS tree ("macos" in this case) into the jar file (not just a single library)
    - The NativeLibraries class copies and loads the JNI and LiteCore libraries separately: The former by architecture, the latter from "universal"
    
A nifty side benefit from this is that, because the native libraries are loaded individually, each goes in its own directory: we no longer need to compute the hash of hashes.

The next step will be convincing Gradle to build the aarch64 JNI lib.

### DIFF
--- a/java/main/java/com/couchbase/lite/internal/NativeLibrary.java
+++ b/java/main/java/com/couchbase/lite/internal/NativeLibrary.java
@@ -17,24 +17,23 @@ package com.couchbase.lite.internal;
 
 import androidx.annotation.NonNull;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
-import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 
 /**
- * For extracting and loading native libraries for couchbase-lite-java.
+ * Extract and load the native libraries.
  */
 final class NativeLibrary {
     private NativeLibrary() { }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: java/main/java/com/couchbase/lite/internal/NativeLibrary.java [hunk 1]
```diff
@@ -43,132 +42,162 @@ private NativeLibrary() { }
...
     @NonNull
     private static String getOsDir(@NonNull String osName) {
         final String os = osName.toLowerCase(Locale.getDefault());
-        if (os.contains("mac")) { return LIB_DIR_MAC; }
-        if (os.contains(LIB_DIR_LINUX)) { return LIB_DIR_LINUX; }
-        if (os.contains(LIB_DIR_WINDOWS)) { return LIB_DIR_WINDOWS; }
-        throw new IllegalStateException("Unrecongnized OS: " + osName);
+        if (os.contains(OS_WINDOWS)) { return OS_WINDOWS; }
+        if (os.contains(OS_LINUX)) { return OS_LINUX; }
+        if (os.contains("mac")) { return OS_MAC; }
```
borrrden, line 146: Curious why not contains(OS_MAC) ?

borrrden (approved):

java/main/java/com/couchbase/lite/internal/NativeLibrary.java
bmeike: Maybe that's better.  Lemme confirm.

pasin (approved):

java/main/java/com/couchbase/lite/internal/NativeLibrary.java
bmeike: That won't work.  The Java recommended way of identifying the MacOS is `System.getProperty("os.name").toLowerCase().contains("mac")`.  The directory containing the MacOS libraries is named "macos" (which is the value of OS_MAC).  I'm going to take the sense of your point and introduce a new constant, MAC_OS_DIR, which will contains the name of the directory.  OS_MAC with be "mac".

PR #186, opened 2023-05-25
CBL-4193: Enable Preemptive Basic Authorization

bmeike:
OkHTTP expects a server that wants Auth info, but does not get it, to respond with a challenge.  Under some conditions (anonymous auth) the SGW will simply close the connection, if it does not get the materials.

The solution is to force the Auth header into the first request.

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
+++ b/common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
@@ -52,7 +52,6 @@ import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLProtocolException;
 import javax.net.ssl.TrustManager;
 
-import okhttp3.Authenticator;
 import okhttp3.Challenge;
 import okhttp3.Cookie;
 import okhttp3.CookieJar;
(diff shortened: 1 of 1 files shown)


borrrden (approved):

common/main/java/com/couchbase/lite/internal/replicator/AbstractCBLWebSocket.java
```diff
@@ -653,23 +650,39 @@ private Request newRequest() {
...
+        if (!(username instanceof String)) { return; }
         final Object password = auth.get(C4Replicator.REPLICATOR_AUTH_PASSWORD);
-        if (!(password instanceof String)) { return null; }
-
-        return (route, response) -> authenticate(response, (String) username, (String) password);
+        if (!(password instanceof String)) { return; }
+
+        final String credentials = Credentials.basic((String) username, (String) password);
+
+        builder.authenticator((route, response) -> authenticate(response, credentials));
```
pasin, line 670: Is this still needed as the auth is already included in the header? I understand that the challenge will not be called again if the auth fails.

bmeike: I am absolutely not certain....  It seems correct to respond to a challenge, if I ever get it.  If I don't ever get it... 🤷🏻

pasin (approved):

PR #188, opened 2023-06-01
Fiddly bits

bmeike:
Replace expired test cert
Stop running URLEndpointListenerTests on a Nexus 5


borrrden (approved):

common/main/java/com/couchbase/lite/ResultSet.java
```diff
@@ -114,7 +114,7 @@ else if (!c4enum.next()) {
         }
 
         // Log outside the the synchronized block
-        Log.w(DOMAIN, msg, err);
+        Log.i(DOMAIN, msg, err);
```
pasin, line 117: Shouldn't be warn or error level?

bmeike: I think it shouldn't be there at all.  I was going to ask you if I could remove it all together.  It is entirely reasonable, it seems to me, to run this thing until it returns `null`.

PR #190, opened 2023-06-02
Normalize OS identification.

bmeike:
The last change broke Windows: the check for the windows OS that exempted it from the attempt to set permissions on the extracted library was wrong, the lib didn't load, and everything failed.

This change normalizes the check for os type so that it is always done the same way, and explains why it is done this way.


borrrden (approved):
Oh dear

PR #194, opened 2023-06-22
CBL-4435:Replicator.close() stops status updates

bmeike:
An extensive comment describes the choice I mede for implementation.

tl; dr: I believe the intention of the close() method is to do exactly what would happen if the GC got the object, but to do it right now.  That is exactly what this implementation does.  It tells LiteCore to close the replicator and then completely forgets about it.  No zombie objects, no long delays, no unexpected exceptions.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -118,6 +118,10 @@ public abstract class AbstractReplicator extends BaseReplicator
     @GuardedBy("getReplicatorLock()")
     private ReplicatorStatus status = ReplicatorStatus.INIT;
 
+    // This is pretty arbitrary.  There's actually no reason that one couldn't restert a closed replicator
+    @GuardedBy("getReplicatorLock()")
+    private boolean closed;
+
     @GuardedBy("getReplicatorLock()")
     @Nullable
     private CouchbaseLiteException lastError;
@@ -170,11 +174,11 @@ public abstract class AbstractReplicator extends BaseReplicator
 
         getDatabase().addActiveReplicator(this);
 
-        final C4Replicator repl = getOrCreateC4Replicator();
+        final C4Replicator c4Repl = getOrCreateC4Replicator();
         synchronized (getReplicatorLock()) {
-            repl.start(resetCheckpoint);
+            c4Repl.start(resetCheckpoint);
 
-            C4ReplicatorStatus status = repl.getStatus();
+            C4ReplicatorStatus status = c4Repl.getStatus();
             if (status == null) {
                 status = new C4ReplicatorStatus(
                     C4ReplicatorStatus.ActivityLevel.STOPPED,
@@ -184,7 +188,7 @@ public abstract class AbstractReplicator extends BaseReplicator
 
             status = updateStatus(status);
 
-            dispatchStatusChange(repl, status);
+            dispatchStatusChange(c4Repl, status);
         }
     }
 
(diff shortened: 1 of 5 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -382,15 +386,43 @@ public void removeChangeListener(@NonNull ListenerToken token) {
...
+    // Replicators *do*, however, dominate a LiteCore object that must be freed explicitly.
+    // Making them, transitively, Autoclosable seems correct.
+    // #2 seems to put an excessive burden on client code: it is a bomb that explodes unless you
+    // wait around and pay very close attention to the state of every replicator.
+    // I dislike #1 because it puts the replicator into an odd state in which it is still kinda
+    // half alive.  The client closed it but it isn't quite gone yet.
+    // I have chosen #3.  My rational is that, if you create and start a replicator and then null out
+    // all references to it, this is exactly what will happen: the replicator will be forgotten and
+    // eventually, GCed, told to stop, and freed.  This method does precisely that: best effort to
+    // release the resources and then forget about them.
```
pasin, line 406: My impression that close() method has synchronize like .NET's dispose() or stream.close(). So when calling close(), the replicator will be stopped and its resource will be released. So maybe #1 with synchronous behavior to ensure that the replicator is stopped?

 Also, I think the Java and .NET could be aligned on the behavior and implementation. CC: @borrrden.

common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -382,15 +386,43 @@ public void removeChangeListener(@NonNull ListenerToken token) {
...
+                listeners = new HashSet<>(changeListeners);
+                newStatus = new ReplicatorStatus(ReplicatorActivityLevel.STOPPED, status.getProgress(), null);
+                status = newStatus;
+            }
+        }
+
         super.close();
+
+        if (newStatus == null) { return; }
+        postChange(true, new ReplicatorChange((Replicator) this, newStatus), listeners);
```
pasin, line 425: I guessed super.close() will just release the LiteCore's replicator. Could this cause a crash if replicator is running? 

I'm not sure about the behavior here when notifying the STOPPED, but the replicator may not be stopped yet? Will it be possible to get notified with BUSY or something else after STOPPED is notified.

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: I see.  Sure, I could do that: no zombie state.  It might take minutes, though.  I have three arguments against that idea:
1) I think Java devs would find it quite surprising if a `close()` method took minutes to run.  Not a blocker, but surprising.
2) There is no way that we could wait for the close, in the finalizer.  That would be catastrophic.  If close is going to do what GC does (which I claim it should), it should not wait.
3) the reason you would wait is in case that the replicator did not close cleanly.  But what would client code do with that information?  I submit that the best effort attempt will usually work, and that when it doesn't work, there's nothing that the user can do, anyway.

The only argument I can see for waiting is that client code wants to do something that they can only do, once the replicator is definitely stopped.  There is a perfectly straightforward way to do that, right now without the `close` method: attach a listener and wait for the stopped state.

common/main/java/com/couchbase/lite/AbstractReplicator.java
bmeike: I have two reasons to believe that it won't:
1) I believe that LiteCore hold references to the replicator.  I believe I just decrement the ref count.  It would be good to confirm this.
2) All replicators are currently released in exactly the way that this close method releases them.  It's been this way since 3.0.  If this way of releasing them might cause crashes, I think we'd have seen them by now.

bmeike: The replicator is, absolutely, stopped, as far as Java is concerned.  It no longer holds any references to the LiteCore replicator object, and the LiteCore object has no valid references to the Java objects.  If the LiteCore replicator attempts to report any status change (or anything like that) to the Java code, the attempt will simply be ignored.

bmeike:
Added JavaDoc comments, as recommended by @pasin , and also an `isClosed()` method, that returns true iff the replicator is closed.

bmeike:
G. Blake Meike
Hey Jens…. Sorry to bug you.
What happens if I c4repl_free (release) a Replicator that is not stopped.  Like, maybe continuous…

Jens Alfke
Should be OK. The replicator keeps a reference to itself as long as it’s active, so it won’t be deleted.

borrrden (commented):
Is there a test for closing a running replicator, and then closing a database to make sure this doesn't cause a close hang?

pasin (approved):

PR #200, opened 2023-07-26
Add public version id

bmeike:


borrrden (approved):
"also uses the new C4 API for query"

PR #201, opened 2023-07-27
Use Flow's CoroutineContext's dispatcher as listener Executor

jeffdgr8:
If no explicit `Executor` is provided to listener `Flow`s, use the `Flow`'s `CoroutineDispatcher` as the listener `Executor`.


bmeike:
https://issues.couchbase.com/browse/CBL-4828

PR #204, opened 2023-08-16
Use Flow's CoroutineContext's dispatcher as the default listener Executor

bmeike:
From Jeff Lockhart

There is a ticket ( CBL-4833 ) to verify the use of the OptIn, before release.


common/main/kotlin/com/couchbase/lite/CommonFlows.kt
```diff
@@ -13,77 +13,94 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// !!! REVIEW BEFORE RELEASE
```
borrrden, line 16: What is this all about?

borrrden (approved):

PR #205, opened 2023-08-25
CBL-4841: Logic bug in Conflict Resolution

bmeike:
Also: CBL-4837: Make the client task pool size more reasonable

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicator.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicator.java
@@ -766,7 +766,7 @@ public abstract class AbstractReplicator extends BaseReplicator
             final ReplicatedDocument rDoc
                 = new ReplicatedDocument(coll.scope, coll.name, docEnd.docId, docEnd.flags, err);
 
-            if (pushing && !CouchbaseLiteException.isConflict(err)) {
+            if (pushing || !CouchbaseLiteException.isConflict(err)) {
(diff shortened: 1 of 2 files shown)


common/main/java/com/couchbase/lite/AbstractReplicator.java
```diff
@@ -766,7 +766,7 @@ private void documentsEnded(@NonNull List<C4DocumentEnded> docEnds, boolean push
             final ReplicatedDocument rDoc
                 = new ReplicatedDocument(coll.scope, coll.name, docEnd.docId, docEnd.flags, err);
 
-            if (pushing && !CouchbaseLiteException.isConflict(err)) {
```
bmeike, line 769: Duh.  This cause the conflict resolver to be called for nearly every document

common/main/java/com/couchbase/lite/internal/exec/ClientTask.java
```diff
@@ -31,18 +31,14 @@
 
 /**
  * Synchronous safe execution of a client task.
- * Motto: Their failure is not our failure.
  * <p>
- * I think it is ok to permit this executor pool to get very large: it is used only
```
bmeike, line 36: Turns out the sentiment expressed in this comment is completely false.  Making the pool that large has no advantage and may cause an OOM creating all those threads.

pasin (approved):

PR #207, opened 2023-08-26
CBL-1048: char[] encoding in Fleece encoder

bmeike:
There are some other "in passing" fixes here.  Mostly though, I'd appreciate having you look at `jstringSlice::jstringSlice(JNIEnv *env, jbyteArray jchars)`

That's the heart of the matter.  `testBasicAuthOptions` verifies that the authenticator password option does, in fact, make it to the network code as a string.

### DIFF
--- a/common/main/cpp/native_glue.cc
+++ b/common/main/cpp/native_glue.cc
@@ -237,7 +237,10 @@ namespace litecore {
             }
         }
 
-        jstringSlice::jstringSlice(JNIEnv *env, jbyteArray jchars) {
+        jstringSlice::jstringSlice(JNIEnv *env, jbyteArray jchars)
+                : _env(env)
+                , _jbytes(jchars) {
+            assert(env != nullptr);
             if (jchars == nullptr) {
                 _slice = kFLSliceNull;
             } else {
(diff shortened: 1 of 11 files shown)


common/main/cpp/native_glue.cc
```diff
@@ -237,6 +237,20 @@ namespace litecore {
...
 
+        jstringSlice::jstringSlice(JNIEnv *env, jbyteArray jchars) {
+            if (jchars == nullptr) {
+                _slice = kFLSliceNull;
+            } else {
+                size_t length = env->GetArrayLength(jchars);
+                if (length <= 0) {
+                    _slice = kFLSliceNull;
+                } else {
+                    auto *chars = env->GetByteArrayElements(jchars, nullptr);
```
pasin, line 248: char* needs be released after use. `jbyteArraySlice` does handle that.

pasin (changes requested):

common/main/cpp/native_glue.cc
bmeike: Woo.... Thanks Pasin!

bmeike: Fixed.  Actually, you fixed it...

pasin (approved):

PR #209, opened 2023-08-29
Unit test for CBL-4841

bmeike:
This is a unit test for CBL-4841.
A logic bug caused any document replicated in a pull replication to be scheduled for conflict resolution.


borrrden:
I don't think that this ticket has enough information inside of it.  It's the same argument as not writing a link only answer in Stack Overflow (if the link is ever gone for whatever reason, you lose context about what happened).

bmeike:
You seem to be saying that this test, even with its copious comments and reference to the bug for which it is a test, should not be part of the codebase.  I am honestly speechless.

Here it is in Kotlin.  See if that's clearer.

pasin (approved):

borrrden:
I said no such thing, I am just speaking regarding the comments on the commit itself.  In my opinion it should contain information on what CBL-4841 is and what the test tests to fix it.

bmeike:
Ahh! I see.  I completely misunderstood what you were talking about.

Fixing now.

PR #214, opened 2023-09-14
CBL-3110: Refactor FLEncoder to use a JNI proxy

bmeike:
My apologies.... this is a sprawling commit.  In addition to the change mentioned in the title, it contains a few other things:

- Better documentation of Android conscrypt support (CBLTrustManager)
- Native FLEncoder.finishJSON *DOES* throw an exception.  Plumb it into Java.  It is not particularly well handled but we'd have to change the API to fix that
- Log old style blobs, if we ever encounter them.
- The original ticket was about adding a proxy to JSONEncoder.  Instead I've just made it a special kind of FLEncoder.
- Add a test for the recently added FLEncoder.writeString(char[])

Everything else is about pushing FLEncoder native code into a proxy.  It is textually a lot of edits but doesn't actually change much


common/main/java/com/couchbase/lite/internal/fleece/FLEncoder.java
```diff
@@ -228,76 +287,35 @@ public boolean writeValue(@Nullable Object value) {
     }
 
     public boolean write(@Nullable Map<String, Object> map) {
-        if (map == null) { beginDict(0); }
+        boolean ok;
+        if (map == null) { ok = beginDict(0); }
         else {
-            beginDict(map.size());
+            ok = beginDict(map.size());
```
pasin, line 293: If it's not ok, should it return right away?

bmeike: The short-circuit `&&` has the same effect.  I was of the understanding that I had to call `endDict` even if `beginDict` failed, thus the way the call to `endDict` happens before the `&&`

bmeike: Fixed so that endDict will not be called if beginDict does not succeed
Same fix for begin and end Array

pasin (approved):

PR #220, opened 2023-10-11
Some refactoring to simplify a few things

bmeike:
Rename the logger: it used to be in a package called "support", which, I think, was meant to hold a bunch of utility stuff.  Logging was the only thing there, though, so I've moved it to a package called "logging" for clarity
Move the Listener interface: Similarly, the listener interface obviously belongs in the listener package.  Put it there
Convert C4CollectionDocObserver and C4CollectionObserver to TaggedWeakPeerBinding:  TaggedWeakPeerBinding is more explicit that peer binding and should be used wherever possible. Only Sockets still use the latter.

All tests pass on multiple devices


common/main/cpp/native_c4observer.cc
```diff
@@ -119,36 +119,38 @@ bool litecore::jni::initC4Observer(JNIEnv *env) {
...
 }
 
 /**
  * Callback method from LiteCore C4DocumentObserver
- * @param docID
- * @param seq
+ * @param ign1 ref for the C4Document observer: ignored
+ * @param ign2 ref for the C4Collection: ignored
+ * @param ign3 the doc id: ignored
+ * @param context the token bound to the observer instance.
```
pasin, line 143: There are 5 params. docID is mixed up with ign3 in the comment above.

common/main/java/com/couchbase/lite/internal/listener/Listenable.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.concurrent.Executor;
+
+import com.couchbase.lite.ChangeListener;
+import com.couchbase.lite.CouchbaseLiteException;
+import com.couchbase.lite.ListenerToken;
+
+
+// Requiring implementer to specify the types of both T and the ChangeListener for T
+// is pretty ugly and annoying.  I could not figure out another way to do it, though...
+public interface Listenable<T, L extends ChangeListener<T>> {
```
pasin, line 30: I might have missed. I don't see it is implemented anywhere in this PR.

pasin (approved):

common/main/java/com/couchbase/lite/internal/listener/Listenable.java
bmeike: wow.... good catch!!

I have no idea what happened there....

PR #223, opened 2023-10-26
Prevent native crashes on closing a ResultSet before referencing its Results.

bmeike:


bmeike:
Tests are passing, except the known Nexus 5 tests and load tests

pasin (approved):

PR #238, opened 2023-12-19
CBL-5269: Make sure that QueryObs functions are called correctly

bmeike:
I removed that Java call to getEnumerator and am now passing results and any error in the callback.

Also added a call to enable(false) in the free method, to guarantee that only the thread-safe version gets called.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractQuery.java
+++ b/common/main/java/com/couchbase/lite/AbstractQuery.java
@@ -65,7 +65,7 @@ abstract class AbstractQuery implements Listenable<QueryChange, QueryChangeListe
         public void start(@NonNull ChangeListenerToken<QueryChange> token) {
             synchronized (liveQueries) {
                 final C4QueryObserver observer = liveQueries.get(token);
-                if (observer != null) { observer.setEnabled(true); }
+                if (observer != null) { observer.enable(); }
             }
         }
 
--- a/common/test/java/com/couchbase/lite/internal/core/C4QueryObserverTest.kt
+++ b/common/test/java/com/couchbase/lite/internal/core/C4QueryObserverTest.kt
@@ -11,11 +11,16 @@
 package com.couchbase.lite.internal.core
 
 import com.couchbase.lite.AbstractIndex
+import com.couchbase.lite.internal.core.impl.NativeC4QueryObserver
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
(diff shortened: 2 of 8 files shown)


common/main/java/com/couchbase/lite/AbstractQuery.java
```diff
@@ -65,7 +65,7 @@ public void put(@NonNull ListenerToken token, @NonNull C4QueryObserver observer)
         public void start(@NonNull ChangeListenerToken<QueryChange> token) {
             synchronized (liveQueries) {
                 final C4QueryObserver observer = liveQueries.get(token);
-                if (observer != null) { observer.setEnabled(true); }
+                if (observer != null) { observer.enable(); }
```
pasin, line 68: This doesn't seem be guard by database exclusive lock. I might have missed something.

common/test/java/com/couchbase/lite/internal/core/C4QueryObserverTest.kt
```diff
@@ -67,10 +71,30 @@ class C4QueryObserverTest : C4BaseTest() {
...
+        val nativeImpl = NativeC4QueryObserver()
+        val latch = CountDownLatch(10)
+        val barrier = CyclicBarrier(10)
+        (0..9).map {
+            Thread {
+                val q = C4Query.create(c4Database, AbstractIndex.QueryLanguage.JSON, query)
+                barrier.await()
+                val obs = C4QueryObserver.create(nativeImpl, { mockQEnum }, q, { _, _ -> })
+                barrier.await()
+                obs.close()
```
pasin, line 93: I just realized that obs.enable() should be called as well. The create alone is thread safe but enabling is not.

common/main/java/com/couchbase/lite/AbstractQuery.java
bmeike: The caller seizes the db lock.  It is already held when this method is called

common/test/java/com/couchbase/lite/internal/core/C4QueryObserverTest.kt
bmeike: Right... Right.  Done

common/main/java/com/couchbase/lite/AbstractQuery.java
pasin: I see.

pasin (approved):

PR #239, opened 2023-12-19
CBL-5269: Make sure calles to c4queryobs_* are thread safe

bmeike:
Minimal changes to assure that  c4queryobs_* are being called correctly


bmeike:
This is the wrong idea

PR #242, opened 2024-01-10
Iterator concurrent modification tests

bmeike:
First steps in the process of normalizing the behavior of collection iterators, across platforms.

In creating this group of tests I've discovered that it makes a difference whether the call to toMutable is on the container or its element.  Will have to add more tests to handle these cases.

testConcurrentModOfSavedArrayIteratorMember fails.  I believe this to be a bug in the Iterator, not the test.

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
+++ b/common/main/java/com/couchbase/lite/internal/core/C4QueryObserver.java
@@ -55,7 +55,7 @@ public class C4QueryObserver extends C4NativePeer {
     }
 
     @NonNull
-    public static C4QueryObserver create(
+    static C4QueryObserver create(
         @NonNull C4QueryObserver.NativeImpl impl,
         @NonNull Fn.Function<Long, C4QueryEnumerator> queryEnumeratorFactory,
         @NonNull C4Query query,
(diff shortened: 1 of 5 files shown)


borrrden (approved):

jeffdgr8 (commented):
These tests look comprehensive to cover the expectation that direct mutations to a collection during iteration should throw `ConcurrentModificationExceptions`, while mutations to child collections should not cause an exception (where currently the first mutation to a saved child collection does).

These should be able to replace [these existing tests](https://github.com/couchbase/couchbase-lite-java-common/blob/04ca16f699b9a955e9e97e3f5306c2e9e253a15a/common/test/java/com/couchbase/lite/ArrayTest.java#L1016-L1054). And adding similar tests for `MutableDictionary` could replace [these as well](https://github.com/couchbase/couchbase-lite-java-common/blob/04ca16f699b9a955e9e97e3f5306c2e9e253a15a/common/test/java/com/couchbase/lite/DictionaryTest.java#L343-L374).

Then there's the question of whether this behavior should also apply to `MutableDocument`. I would think it should to be consistent, but currently there aren't similar tests and `ConcurrentModificationExceptions` isn't thrown by its iterator.

bmeike:
Thanks @jeffdgr8. Excellent.
- Will remove the redundant tests.
- Will add tests for Dictionary and remove redundant tests
- Will add tests for Dictionary.

bmeike:
Done, FWIW.
It is amazing how painful it is to write 39 nearly identical tests....

jeffdgr8 (approved):
Of course there's also the fact there's a couple dozen ways to mutate each collection type, but I'm not sure it's worth blowing up the tests exponentially to account for each of them. 😅

PR #244, opened 2024-01-15
Pr/cbl 5310

bmeike:
This should really be two separate commits.  Sorry about that.
The flaky test thing is just that: a couple of simple changes that allow specifying a list of expected exceptions instead of just a single one.

The big change is CBL-5310 which gets ConcurrentModification exceptions working correctly for all three Container types: Document, Dictionary, Array

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/fleece/MValue.java
+++ b/common/main/java/com/couchbase/lite/internal/fleece/MValue.java
@@ -18,8 +18,6 @@ package com.couchbase.lite.internal.fleece;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import com.couchbase.lite.MValueConverter;
 import com.couchbase.lite.internal.utils.Preconditions;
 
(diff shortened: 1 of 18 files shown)


common/main/java/com/couchbase/lite/internal/fleece/MValue.java
```diff
@@ -46,6 +44,8 @@
  * <p>
  * The regrettable upside-down dependency on MValueConverter provides access to package
  * visible symbols in com.couchbase.lite.
+ * <p>
+ * It worries me that this isn't thread safe... but, as I say, I've never seen it be a problem.
```
pasin, line 48: The synchronization happens in MutableArray and MutableDictionary so for most part, I think it would be OK. One skeptical part which is when setting mutate to the parent object may need to look through if a synchronization is needed there.

pasin (approved):

common/main/java/com/couchbase/lite/internal/fleece/MValue.java
bmeike: Ah... I see.
I think this is might possibly be the mistaken belief that you only need to synchronize writes.  That's not how Java works: "If multiple threads access the same mutable state variable without synchronization, your program is broken." -- Brian Goetz

PR #245, opened 2024-01-16
Database.getDefaultCollection() should be non-null

jeffdgr8:
Translate default scope/collection tests from Swift. Also, add missing `@Test` annotation to some collection tests.


bmeike:
Jeff: our API tzar wants to keep this nullable.  Being unable to delete the default collection is kind of an odd artifact of the current implementation.  We may, very well, allow deleting it at some point in the future.

jeffdgr8:
[The Swift API is non-null.](https://github.com/couchbase/couchbase-lite-ios/blob/d491d1398a48054a40b8d119d964316ea191c051/Swift/Database.swift#L491) Should this be changed to nullable then?

bmeike:
Well... they should be the same!  I will check.

bmeike:
Turns out that *all* of the other platforms went with non-null.  It is easier to change Java than to change all the rest.  This will go into 3.2.0 and subsequent releases.   3.1.7 is already out the door, so I won't be able to do it there.

bmeike:
Turns out that all the other platforms went with non-null, and it is easier to change Java than to change all the others.
I will make that method @NonNull in 3.2 and subsequent releases.

bmeike:
Jeff: this change is now in the master and release/3.2 branches.  Thanks.

PR #246, opened 2024-01-25
CBL-5323: Vector Index API

bmeike:
Appears to work:  Attempt to use it fails with the expected error: Missing SQLite library


common/main/java/com/couchbase/lite/internal/utils/Preconditions.java
```diff
@@ -75,6 +75,13 @@ public static int assertNotZero(int n, @NonNull String name) {
         return n;
     }
 
+    public static long assertUnsigned(long n, @NonNull String name) {
+        if ((n < 0) || (n >= (1L << 33))) {
```
borrrden, line 79: I'm guessing this is a range check for 32-bit uint?  If so it needs to check >= 1L << 32, not 33.  The max value of uint is 2^32 - 1

borrrden (changes requested):

borrrden (approved):

PR #248, opened 2024-01-30
Update VS to match new LiteCore API

bmeike:
Big change at the last minute....

### DIFF
--- a/common/main/cpp/native_c4collection.cc
+++ b/common/main/cpp/native_c4collection.cc
@@ -309,7 +309,7 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Collection_createPredictiveIn
 /*
  * Class:     com_couchbase_lite_internal_core_impl_NativeC4Collection
  * Method:    createVectorIndex
- * Signature: (JLjava/lang/String;Ljava/lang/String;IJIIIIJJ)V
+ * Signature: (JLjava/lang/String;Ljava/lang/String;JIJIJJJJ)V
  */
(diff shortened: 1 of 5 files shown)


common/main/cpp/native_c4collection.cc
```diff
@@ -318,29 +318,29 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Collection_createVectorIndex(
...
-
-    options.vector.numCentroids = centroids;
-    options.vector.metric = (C4VectorMetric) metric;
-    options.vector.minTrainingSize = minTrainingSize;
-    options.vector.maxTrainingSize = maxTrainingSize;
+    options.vector = {};
+    options.vector.dimensions = (unsigned) dimensions;
+    options.vector.metric = (C4VectorMetricType) metric;
+    options.vector.clustering = {kC4VectorClusteringFlat, (unsigned) centroids, 0, 0};
+    options.vector.encoding = { (C4VectorEncodingType) encoding, (unsigned) subquantizers, (unsigned) bits };
```
pasin, line 335: FYI : C4VectorEncodingType has the first item in the enum as kC4VectorEncodingDefault so need to make sure that the encoding value passed from Java has offset that.

pasin (approved):

common/main/cpp/native_c4collection.cc
borrrden: A weird choice when it could just be aliased to another value in LiteCore but good to know...

PR #251, opened 2024-02-13
Working VectorSearch

bmeike:
Update tests to accomodate date formats in LiteCore 3.2.0-137
Clean up some other test stuff.

### DIFF
--- a/java/main/java/com/couchbase/lite/internal/NativeLibrary.java
+++ b/java/main/java/com/couchbase/lite/internal/NativeLibrary.java
@@ -16,6 +16,7 @@
 package com.couchbase.lite.internal;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -61,7 +62,7 @@ final class NativeLibrary {
 
     private static final String DIGEST_MD5 = ".MD5";
 
-    private static final AtomicBoolean LOADED = new AtomicBoolean(false);
+    private static final AtomicBoolean LOADED = new AtomicBoolean();
 
     /**
      * Extracts the two native libraries from the jar file, puts them on the file system and loads them.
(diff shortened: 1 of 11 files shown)


java/main/java/com/couchbase/lite/internal/NativeLibrary.java
```diff
@@ -177,6 +197,14 @@ private static String getJniArchDir(@NonNull String rootPath, @NonNull String os
         throw new IllegalStateException("Unsupported JNI architecture: " + osName + "/" + archName);
     }
 
+    @NonNull
+    private static String getLibExtension(@NonNull String osName) {
+        if (isWindows(osName)) { return ".dll"; }
+        if (isMacOS(osName)) { return ".dylib"; }
```
borrrden, line 203: This is not actually going to be correct because we don't distribute the native lib as a dylib right now.  We do for LiteCore but vector search is built as an xcframework so the extension will be empty.  If you want a dylib instead this is going to require another build target separate from the "apple" one.  OR maybe this will work, you could copy the binary blob out of the framework and add the dylib extension at download time.

borrrden (approved):
Will discuss more offline about how to handle macOS.

java/main/java/com/couchbase/lite/internal/NativeLibrary.java
bmeike: It has been loading the thing named "CouchbaseLiteVectorSearch.dylib" correctly.  Perhaps that is because it thinks that is its full name?

pasin: @borrrden I might have given you inaccurate info about macOS binary. I just realized that CBL-C currently releases dylib for macOS. The stucture of the release zip file is like :

```
.
├── LICENSE.txt
├── include
├── lib
│   ├── cmake
│   │   └── CouchbaseLite
│   │       ├── CouchbaseLiteConfig.cmake
│   │       ├── CouchbaseLiteConfigVersion.cmake
│   │       ├── CouchbaseLiteTargets-minsizerel.cmake
│   │       └── CouchbaseLiteTargets.cmake
│   ├── libcblite.3.1.3.dylib
│   ├── libcblite.3.dylib -> libcblite.3.1.3.dylib
│   └── libcblite.dylib -> libcblite.3.dylib
└── notices.txt
```

pasin (approved):

PR #260, opened 2024-03-12
Android/vf/3.1.6 vf001

bmeike:
This is the VF for CBSE-16470
I would very much appreciate additional eyes on it.  It does three things:

1. Native JSONEncoder_finishJSON now throws a LiteCore exception instead of returning null.  All of the methods that use it wrap the (checked) exception in (unchecked) IllegalStateException.  No API change; Container toJSON methods cannot return null.
1. The context object that becomes the parent of the tree of children belonging to a ResultSet, now holds a reference to the ResultSet.  There will be references to the ResultSet until the last child goes away
1. The close() method on ResultSet no longer does anything.  Instead the Fleece object is freed by a finalizer.  The finalizer code is insane because Android does not appear to handle the use of locks correctly.

### DIFF
--- a/common/main/java/com/couchbase/lite/Result.java
+++ b/common/main/java/com/couchbase/lite/Result.java
@@ -26,7 +26,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
-import com.couchbase.lite.internal.DbContext;
 import com.couchbase.lite.internal.core.C4QueryEnumerator;
 import com.couchbase.lite.internal.fleece.FLArrayIterator;
 import com.couchbase.lite.internal.fleece.FLValue;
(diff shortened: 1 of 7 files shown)


common/main/java/com/couchbase/lite/Result.java
```diff
@@ -569,7 +574,7 @@ private void assertInBounds(int index) {
     }
 
     private void assertOpen() {
-        if (rs.isClosed()) {
+        if (context.getResultSet().isClosed()) {
```
borrrden, line 577: Hmmm, is this ever going to be true?  Doesn't the no-op close prevent this?

borrrden (approved):
I think the goal is accomplished, looks like one place now is somewhat of a no-op (marked with comment)

common/main/java/com/couchbase/lite/Result.java
bmeike: yeah.   good point.

bmeike:
Thanks for the review, Jim!

PR #275, opened 2024-04-19
CBL-5572: ClientTask executor should not fail on RejectedExecution

bmeike:
CBL-5574: Less alarming message on ClientTask executor failure
CBL-5573: Replace the SynchronousQueue on the ClientTaskExecutor
Add Expression.in(List<Expression>)
Add "test" for QueryObserver issue


pasin (approved):
Java has many types of queue to choose from. I think it's a good idea to get the queue bound. I don't know what the right size should be but if users don't block the queue for too long, this config seems to be enough.

PR #279, opened 2024-04-23
CBL-5655: Native crash using ResultSet component after close

bmeike:
I so wish I could have come up with a better solution.  This just turns a native crash into a Java crash with, perhaps, a better error message.

Big change for a release branch.  It was mostly tested in a VF.

### DIFF
--- a/common/main/java/com/couchbase/lite/ResultSet.java
+++ b/common/main/java/com/couchbase/lite/ResultSet.java
@@ -25,7 +25,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
-import com.couchbase.lite.internal.DbContext;
 import com.couchbase.lite.internal.core.C4QueryEnumerator;
 import com.couchbase.lite.internal.support.Log;
 import com.couchbase.lite.internal.utils.Preconditions;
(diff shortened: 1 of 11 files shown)


common/main/java/com/couchbase/lite/ResultSet.java
```diff
@@ -148,11 +145,30 @@ public List<Result> allResults() {
...
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        try {
+            final C4QueryEnumerator qEnum = c4enum;
+
+            // !!! This absolutely should hold the db lock the way close() does.
+            // If it does, though, it stands a good chance of timing out the finalizer thread.
+            // Cross your fingers and pray.
```
pasin, line 166: This comment scares me. Sound like a time bomb for a CBSE that is difficult to know what is going on. 

@borrrden Have you mentioed before that .NET has never locked this and never seen any issues?

bmeike: I think we have to do this.  If we try to seize the lock, we are guaranteed that the finalizer thread will time out.

pasin (approved):

PR #285, opened 2024-04-26
CBL-5683: Database.getDefaultCollection() should be non-null

bmeike:
For compatibility with other platforms.

H/T to @jeffdgr8  for pointing this out and for supplying much of the code.

### DIFF
--- a/common/main/java/com/couchbase/lite/DatabaseChange.java
+++ b/common/main/java/com/couchbase/lite/DatabaseChange.java
@@ -30,13 +30,10 @@ import java.util.List;
 public class DatabaseChange {
     @NonNull
     private static Collection getDefaultCollection(@NonNull Database database) {
-        CouchbaseLiteException fail = null;
-        try {
-            final Collection defaultCollection = database.getDefaultCollection();
-            if (defaultCollection != null) { return defaultCollection; }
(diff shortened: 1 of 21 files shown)


common/main/java/com/couchbase/lite/DatabaseChange.java
```diff
@@ -30,13 +30,10 @@
 public class DatabaseChange {
     @NonNull
     private static Collection getDefaultCollection(@NonNull Database database) {
-        CouchbaseLiteException fail = null;
-        try {
-            final Collection defaultCollection = database.getDefaultCollection();
-            if (defaultCollection != null) { return defaultCollection; }
+        try { return database.getDefaultCollection(); }
+        catch (CouchbaseLiteException e) {
+            throw new CouchbaseLiteError("Database " + database.getName() + " has no default collection", e);
```
borrrden, line 35: Not sure about the wording here given that the default collection cannot be deleted.  How else would it vanish?

borrrden (dismissed):

common/main/java/com/couchbase/lite/DatabaseChange.java
bmeike: Good call.  Fixed

borrrden (approved):

PR #287, opened 2024-04-30
CBL-5691: Client side proxy does not work

bmeike:
Also:
CBL-5689: Don't leak basic auth credentials in logs
Important note on dbConfig.get/setDirectory
Fix spelling error in dbChange exception message


borrrden (approved):
I think this boils down to using `proxyAuthenticator` instead of `authenticator` for proxy authentication

PR #289, opened 2024-05-09
Several fixes

bmeike:
This is bigger than it should be.  Sorry about that.  The native crash is the biggest fix.  The ClientTask fix is more contained but, perhaps, more important.  The rest is just preening.

CBL-5729: Prevent native crash on unknown log domain
CBL-5714: API Doc: Can't start a replicator inside a batch block
CBL-5658: Tune the ClientTask executor
CBL-5651: Tests for LiteCore callbacks

### DIFF
--- a/common/main/cpp/native_c4.cc
+++ b/common/main/cpp/native_c4.cc
@@ -21,6 +21,7 @@
 #include <sys/time.h>
 #endif
 
+#include <unordered_set>
 #include "native_glue.hh"
 #include "com_couchbase_lite_internal_core_impl_NativeC4.h"
 #include "com_couchbase_lite_internal_core_impl_NativeC4Log.h"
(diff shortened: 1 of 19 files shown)


common/main/cpp/native_c4.cc
```diff
@@ -138,6 +139,26 @@ static void logCallback(C4LogDomain domain, C4LogLevel level, const char *fmt, v
...
+ * we can do so at any time, including before Core creates it.  The problem with that is that
+ * LiteCore doesn't keep the names, itself.  Unless the domain already exists, it will use
+ * the string that we pass to it.  That means that string has to be stored somewhere permanent.
+ * It is possible that this can go away when CBL-5726 is resolved
+ */
+namespace litecore {
+    namespace jni {
+        static std::unordered_set<std::string> gDomainNames;
+
+        // Assert: jdomain is not null
```
borrrden, line 154: null would be fine, LiteCore auto maps null to empty string aka default

borrrden (approved):

common/main/cpp/native_c4.cc
bmeike: I think this will fail on null:
`std::string domain = JstringToUTF8(env, jdomain);`

PR #293, opened 2024-05-21
CBL-5678: set state to CLOSING in coreRequestClose

bmeike:
Although the only ticketable change in this PR, it is not the biggest change.  In addition:

- [Workaround for CBL-5780: Listeners cani't be stopped.](https://github.com/couchbase/couchbase-lite-java-common/commit/28c871aae3e9ba0e4aa4f33aa8b5e13678f8fbae)
- Fix a regression caused by test time-out timer that doubled the time that it took to run a test.

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractDatabase.java
+++ b/common/main/java/com/couchbase/lite/AbstractDatabase.java
@@ -1576,19 +1576,34 @@ abstract class AbstractDatabase extends BaseDatabase
         try {
             for (int i = 0; ; i++) {
                 verifyActiveProcesses();
+
                 if ((i >= DB_CLOSE_MAX_RETRIES) && (closeLatch.getCount() > 0)) {
                     throw new CouchbaseLiteException("Shutdown failed", CBLError.Domain.CBLITE, CBLError.Code.BUSY);
                 }
-                if (closeLatch.await(DB_CLOSE_WAIT_SECS, TimeUnit.SECONDS)) { break; }
+
+                if (closeLatch.await(DB_CLOSE_WAIT_SECS, TimeUnit.SECONDS)) {
+                    try {
+                        synchronized (getDbLock()) { onShut.accept(c4Db); }
+                        break;
+                    }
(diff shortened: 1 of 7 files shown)


common/main/java/com/couchbase/lite/AbstractDatabase.java
```diff
@@ -1576,19 +1576,34 @@ private void shutdown(boolean failIfClosed, Fn.ConsumerThrows<C4Database, LiteCo
...
+                    }
+                }
+
+                // If we get here then, despite the fact that it appears to us that all
+                // active processes have been stopped, LiteCore has other ideas.
+                // We have no way of finding out what LiteCore thinks, other than waiting
+                // a bit and trying the again.  Since verifyActiveProcess will count down
+                // the latch, we need a new one with a count of at least 2 in order to force
+                // a wait
+                closeLatch = new CountDownLatch(2);
```
pasin, line 1603: It took me awhile to understand why setting to 2. 

As closeLatch() is an instance variable, could it be a problem for an edge case when database.close() is somehow called twice from a different thread while the first call is still waiting.

bmeike: No.  It can't be called from a different thread.  This point in the shutdown method can be reached only once, for any Database instance.  See this code at the beginning of the method:

`        synchronized (getDbLock()) {
            final boolean open = isOpenLocked();
            Log.d(DOMAIN, "Shutdown (%b, %b)", failIfClosed, open);
            if (!(failIfClosed || open)) { return; }

            c4Db = getOpenC4DbLocked();
            setC4DatabaseLocked(null);
            // mustBeOpen will now fail, which should prevent any new processes from being registered.
            
            ...
`
Only one thread at a time can be in the synchronized section.  That thread will set c4db to null.  Any subsequent thread will exit immediately
The rest of the method will be executed only once per Db

pasin: Got it.

pasin (approved):

PR #294, opened 2024-05-23
Remove getBinaryFileLevel from the public API

bmeike:
Also clean up some ResultSet code


pasin (approved):
I'm only see the change related to getBinaryFileLevel in `jni_android.ld` so I guessed there are some other changes in another repo.

bmeike:
@pasin : exactly.  most of the linker files are platform/ee specific

PR #297, opened 2024-06-06
CBL-5794: FLSliceResult leaks memory

bmeike:
FLSliceResult is not freeing the block of memory it frames.  This is an enormous memory leak.

This is a minimal backport of https://github.com/couchbase/couchbase-lite-java-common/pull/296: the SliceResult if freed by the finalizer; not explicitly.

Also removed an unused file logger function

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/fleece/FLEncoder.java
+++ b/common/main/java/com/couchbase/lite/internal/fleece/FLEncoder.java
@@ -62,6 +62,8 @@ public abstract class FLEncoder extends C4NativePeer {
         @NonNull
         FLSliceResult nFinish2(long peer) throws LiteCoreException;
         @NonNull
+        FLSliceResult nFinish3(long peer) throws LiteCoreException;
+        @NonNull
         String nFinishJSON(long peer) throws LiteCoreException;
         void nFree(long peer);
     }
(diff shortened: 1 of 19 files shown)


common/main/java/com/couchbase/lite/internal/fleece/FLEncoder.java
```diff
@@ -335,6 +337,12 @@ public boolean write(@Nullable List<?> list) {
...
     public byte[] finish() throws LiteCoreException { return impl.nFinish(getPeer()); }
 
+    // NOTE: the FLSliceResult returned by this method must be released by the caller
     @NonNull
     public FLSliceResult finish2() throws LiteCoreException { return impl.nFinish2(getPeer()); }
+
+    // NOTE: We expect the FLSliceResult returned by this method to be handed, immediately,
+    // to someone else (LiteCore) who will release it.  Java does release the memory.
+    @NonNull
+    public FLSliceResult finish3() throws LiteCoreException { return impl.nFinish3(getPeer()); }
```
pasin, line 347: Trying to think about a better name than 1, 2, 3 but it's hard to come up with a good one. Maybe finishAndGetManagedFLSliceResult (how long is this?).

pasin (approved):

PR #299, opened 2024-06-07
Explicit VS enabling

bmeike:
also CBL-5810: Update VS tests to v2.0.1


pasin (commented):
I understand that we will need a new LiteCore API for this. Waiting to discuss info with @borrrden

PR #302, opened 2024-06-13
CBL-5848: Dates cannot be encoded as Parameters

bmeike:


common/main/java/com/couchbase/lite/Blob.java
```diff
@@ -621,8 +622,10 @@ private byte[] getContentFromDatabase() {
         Preconditions.assertNotNull(database, "database");
 
         final byte[] newContent;
-        try (C4BlobStore blobStore = database.getBlobStore(); C4BlobKey key = C4BlobKey.create(blobDigest)) {
-            newContent = blobStore.getContents(key).getContent();
+        try (C4BlobStore blobStore = database.getBlobStore();
+             C4BlobKey key = C4BlobKey.create(blobDigest);
+             FLSliceResult result = blobStore.getContents(key)) {
+            newContent = result.getContent();
```
pasin, line 628: Is the indent slightly off?

pasin (approved):

common/main/java/com/couchbase/lite/Blob.java
bmeike: yep.  Thx.

PR #305, opened 2024-06-18
CBL-5845: Port LocalRef handling from 3.1

bmeike:
This is a pretty big commit.  There are two functional changes:
- Use FLEncoder_GetErrorMessage to give better error messages when the encoder fails
- LocalRefs are explicitly freed as soon as possible (DeleteLocalRef, or DetachCurrentThread)

The rest of the changes (that is, most of them)  are just an attempt to normalize the code:
- logError renamed as jniLog, because it is used to log things other than errors
- pointers are explicitly compared to nullptr for verification
- SliceResults are verified using their bool op
- "auto" used only if the right side of the assignment makes the type absolutely clear.
- FSSliceResults are freed with FLSliceResult_Release; C4Slices are freed with its alias c4slice_free
- boolean returns are now named "ok", pretty much universally.
- some line wrapping to make the code fit on the page

### DIFF
--- a/common/main/cpp/native_c4listener.cc
+++ b/common/main/cpp/native_c4listener.cc
@@ -70,10 +70,11 @@ static bool httpAuthCallback(C4Listener *ignore, C4Slice authHeader, void *conte
     if (getEnvStat == JNI_OK) {
         jstring _header = toJString(env, authHeader);
         res = env->CallStaticBooleanMethod(cls_C4Listener, m_C4Listener_httpAuthCallback, (jlong) context, _header);
-        env->DeleteLocalRef(_header);
+        if (_header) env->DeleteLocalRef(_header);
     } else if (getEnvStat == JNI_EDETACHED) {
         if (attachCurrentThread(&env) == 0) {
-            res = env->CallStaticBooleanMethod(cls_C4Listener,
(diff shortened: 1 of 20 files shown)


common/main/cpp/native_c4listener.cc
```diff
@@ -749,17 +754,23 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4KeyPair_generateSelfSignedCer
     auto keys = (C4KeyPair *) c4KeyPair;
 
     int size = env->GetArrayLength(nameComponents);
-    auto subjectName = new C4CertNameComponent[size];
+    C4CertNameComponent *subjectName = new C4CertNameComponent[size];
```
borrrden, line 757: I see what you are doing with the rejection of auto but it was kind of made for situations like this where the object name shows up twice in one line (comment only, not a review)

borrrden (approved):

pasin (approved):

common/main/cpp/native_c4listener.cc
bmeike: Yeah.... Overzealous.  WIll use "auto" with "new" in the future.

PR #306, opened 2024-06-20
CBL-5844: Port LocalRef handling from 3.1

bmeike:
A minimal, functionality only, backport of the changes just PRed in Boron

### DIFF
--- a/common/main/cpp/native_c4listener.cc
+++ b/common/main/cpp/native_c4listener.cc
@@ -70,10 +70,11 @@ static bool httpAuthCallback(C4Listener *ignore, C4Slice authHeader, void *conte
     if (getEnvStat == JNI_OK) {
         jstring _header = toJString(env, authHeader);
         res = env->CallStaticBooleanMethod(cls_C4Listener, m_C4Listener_httpAuthCallback, (jlong) context, _header);
-        env->DeleteLocalRef(_header);
+        if (_header != nullptr) env->DeleteLocalRef(_header);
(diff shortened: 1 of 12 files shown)


common/main/cpp/native_c4listener.cc
```diff
@@ -177,6 +179,7 @@ static bool doDecryptCallback(
     int n = (int) input.size;
 
     jbyteArray encData = env->NewByteArray(n);
+    if (encData == nullptr) return false;
```
borrrden, line 182: If this ever happens you are probably never going to figure out why without some kind of message in the logs

common/main/cpp/native_c4observer.cc
```diff
@@ -189,15 +189,24 @@ static jobjectArray
 c4DocChangesToJavaArray(JNIEnv *env, C4CollectionChange changes[], uint32_t nChanges, bool external) {
     jobjectArray array = env->NewObjectArray((jsize) nChanges, cls_C4DocChange, nullptr);
     for (size_t i = 0; i < nChanges; i++) {
+        jstring _docId = toJString(env, changes[i].docID);
```
borrrden, line 192: Underscore naming for a local variable?

borrrden (approved):

common/main/cpp/native_c4listener.cc
bmeike: Yeah.... that's a good point.  Will add in a future commit

PR #319, opened 2024-07-23
CBL-6051: Use wildcard type for Mutable container setData

bmeike:
Also fix a warning in LogTest


borrrden (approved):
Assuming this change to public functions is non breaking

PR #320, opened 2024-07-30
CBL-6068: FullSync option

bmeike:
I'm concerned about this, after reviewing Jim's PR.  I concluded that there was no need to consider the full-sync flag in any of the db.copy methods. Perhaps that is not correct?


borrrden:
I can't think of a specific need, but on the other hand why not do it?  If the user doesn't want the flag to be applied during the copy they can simply pass a config with false for the full sync option.

borrrden (commented):
Commented above

borrrden (approved):
Cancelling my previous review.  I forgot that all of this copy stuff happens on a temporary db and any error or power loss would simply result in the database not being copied.  No data loss, corruption, or anything of the sort.

PR #322, opened 2024-07-30
CBL-6121: FullSync option

bmeike:
Identical to the beryllium PR


borrrden:
This is potentially point zero road territory but if you are not using that strategy then fair enough.

borrrden (approved):

PR #328, opened 2024-08-15
CBL-6075: Implement phantom ref driven cleaner

bmeike:


borrrden (approved):
Essentially the TL;DR is that this model allows you to make more threads for finalization so that the one existing one from the runtime doesn't get backed up.  All the details are around making sure that the existing finalizer thread simply marks the object as a zombie and moves on so that the new thread can handle it.

PR #331, opened 2024-08-27
CBL-6135: Audit LiteCore locks

bmeike:
First step: annotate the native methods.


common/main/java/com/couchbase/lite/internal/core/impl/NativeC4Blob.java
```diff
@@ -91,6 +95,7 @@ public int nRead(long peer, byte[] data, int offset, long len) throws LiteCoreEx
     private static native void free(long peer);
 
     // BlobStore
+    @GuardedBy("dbLock")
```
pasin, line 98: Is the dbLock an object defined on the class that use this native method?

pasin (approved):

common/main/java/com/couchbase/lite/internal/core/impl/NativeC4Blob.java
bmeike: No.   I don't think that is relevant, though.  Nobody is checking that string:  It appears to be, essentially, a comment.

FWIW, it is the same name that I use for the dbLock, elsewhere in the code.  In the other places, it is actually the name of the variable holding the lock.

PR #332, opened 2024-09-06
CBL-5176: Implement ArrayIndex API

bmeike:
Almost all the way to LiteCore


borrrden (commented):
There is one interesting situation here that probably should be considered

If the user passes an array with just an empty string, it seems that will turn into a final string of "" to be sent to LiteCore, which is valid (the same as not passing expressions at all) but possibly not what the user wanted to do.  On C# what I will do is ensure there is at least one non-null non-empty string in the list to avoid this particular situation.

bmeike:
The following are all equivalent:
new ArrayIndexConfiguration("<some path>")
new ArrayIndexConfiguration("<some path>", "")
new ArrayIndexConfiguration("<some path>", Arrays.asList(""))

I verify that:
- the list of expressions is not null
- the list of expressions does not contains a null
- the list of expressions is not empty.

If you believe that this is not correct, please suggest an alternative.

borrrden (approved):

PR #333, opened 2024-09-09
Backport Cleaner and ArrayIndex

bmeike:
Also change Date from string conversion to us UTC instead of GMT.  Mostly irrelevant since times should specify Zulu


borrrden (approved):
Assuming this is all stuff that I've already reviewed before and that there were not any hideous conflicts.

pasin (approved):

PR #334, opened 2024-09-09
CBL-6075: Implement phantom ref driven cleaner

bmeike:
I'm pretty happy with this.


borrrden (approved):
Assuming this is not changed much from the code that I reviewed before (or at least attempted to).

bmeike:
Good assumption!  I should have noted that.

PR #335, opened 2024-09-11
CBL-6214: Enable Version Vectors

bmeike:


common/test/java/com/couchbase/lite/internal/core/C4BaseTest.java
```diff
@@ -163,17 +163,13 @@ protected void checkChanges(
     }
 
     protected int getTestDbFlags() {
-        int flags = C4Database.DB_FLAGS;
-        if (C4Database.VERSION_VECTORS_ENABLED) { flags |= C4Constants.DatabaseFlags.FAKE_CLOCK; }
-        return flags;
+        return C4Database.DB_FLAGS | C4Constants.DatabaseFlags.FAKE_CLOCK;
```
borrrden, line 166: The tests should all pass without this flag present.

common/main/java/com/couchbase/lite/internal/core/C4Database.java
```diff
@@ -51,13 +51,11 @@
...
 public abstract class C4Database extends C4NativePeer {
-    public static final boolean VERSION_VECTORS_ENABLED = false;
-
     @VisibleForTesting
-    static final int DB_FLAGS = C4Constants.DatabaseFlags.CREATE;
+    public static final String DB_EXTENSION = ".cblite2";
 
     @VisibleForTesting
-    public static final String DB_EXTENSION = ".cblite2";
+    static final int DB_FLAGS = C4Constants.DatabaseFlags.CREATE;
```
borrrden, line 58: Why are these flags not including versionvector by default?

common/main/java/com/couchbase/lite/internal/core/C4Database.java
```diff
@@ -250,15 +247,13 @@ static C4Database getDatabase(
...
-        if (VERSION_VECTORS_ENABLED) { flags |= C4Constants.DatabaseFlags.VERSION_VECTORS; }
-
         return new ManagedC4Database(
             impl,
             name,
             impl.nOpen(
                 parentDirPath,
                 name,
-                flags,
+                flags | C4Constants.DatabaseFlags.VERSION_VECTORS,
```
borrrden, line 256: This seems odd to me to tack this on here.  What is the flow where this is expected to be missing?

common/main/java/com/couchbase/lite/internal/core/C4Database.java
bmeike: Because the VersionVector flag is added in the code you highlighted below.

common/main/java/com/couchbase/lite/internal/core/C4Database.java
bmeike: If version vectors is on, it is always on.  All databases get created with the version vectors flag.  Putting it here guarantees that.  If it were not here, it would have to be in every place that called this method.  That would not make sense.
There is no other flag that works this way.  There are tests here that do not apply the CREATE flag; that do apply the READ_ONLY flag.  There used to be tests that applied the FAKE_CLOCK and production code that did not.  The VERSION_VECTOR flag is the only one that either must always be on, or always be off.

Your comment makes me think I should get rid of the DB_FLAGS constant.  It is confusing.  I'm going to create a ticket for myself, to do that.  Looking at the code, there is no reason for most of these methods to pass a flags argument around at all.

common/test/java/com/couchbase/lite/internal/core/C4BaseTest.java
bmeike: They do!   Cool!
Removed.

borrrden (approved):

PR #340, opened 2024-09-20
Pre LockManager cleanup

bmeike:
This is just a lot of cleanup of JNI code, before addressing the actual lock audit.
A lot of files touched; not much actual code change

Here are the important changes:
- Catch more leaking PeerBinding tokens: an exception thrown after a call to `reserve` will permanently leak the token
- Move full text match test utils to TestUtils: The entire FullTextMatch interface is used only by tests.
- Add locking annotations to the NativeC4 classes: propagate Jianmin's lock notes as @GuardedBy annotations in the NativeC4 classes
- An attempt to read 0 bytes from a closed Blob will throw an IOException. It used to return 0 and throw only if more than 0 bytes were read.
- An attempt to install a blob in a null db will now throw a CouchbaseLiteError instead of an NPE

!!! Running out of time.  Added an important change to the RevHistory call that gets a fresh copy of the document with `kDocGetAll`


common/main/cpp/native_c4document.cc
```diff
@@ -200,9 +201,21 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Document_getRevisionHistory(
...
 
-    return toJString(env, revHistory);
+    C4Error error{};
+    C4Document* allDoc = c4coll_getDoc((C4Collection*) jcoll, doc->docID, false, kDocGetAll, &error);
+    if ((allDoc == nullptr) && (error.code != 0)) {
+        throwError(env, error);
+        return 0;
+    }
+
+    auto revHistory = c4doc_getRevisionHistory(doc, (unsigned int) maxRevs, backToRevs, nBackToRevs);
```
pasin, line 213: This should use allDoc instead of doc.

pasin (changes requested):

common/main/cpp/native_c4document.cc
bmeike: Whoops.  Right you are.

bmeike: fixed

borrrden (approved):

pasin (approved):

PR #347, opened 2024-10-18
CBL-6296: Check unnest path in ArrayIndex tests

bmeike:
Plumb c4Index_getOptions into the JNI and use it in the ArrayIndex tests

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/core/C4Index.java
+++ b/common/main/java/com/couchbase/lite/internal/core/C4Index.java
@@ -1,0 +1,84 @@
+//
+// Copyright (c) 2024 Couchbase, Inc All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


common/main/java/com/couchbase/lite/internal/core/C4Index.java
```diff
@@ -0,0 +1,84 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
+
+import com.couchbase.lite.LiteCoreException;
+import com.couchbase.lite.LogDomain;
+import com.couchbase.lite.internal.core.impl.NativeC4Index;
+
+
+public final class C4Index extends C4NativePeer {
```
pasin, line 28: Wondering if this is moved from somewhere which was implemented for the lazy index feature?

pasin (approved):

borrrden (approved):

common/main/java/com/couchbase/lite/internal/core/C4Index.java
bmeike: Had to move it so that the tests can find it.  It was only needed for EE before.... but now the tests need to be able to find the native peer...

PR #351, opened 2024-10-24
Add revision history test

bmeike:
Just a couple small tweaks after reviewing the backport to 3.2

### DIFF
--- a/common/test/java/com/couchbase/lite/DocumentTest.java
+++ b/common/test/java/com/couchbase/lite/DocumentTest.java
@@ -2712,6 +2712,16 @@ public class DocumentTest extends BaseDbTest {
             () -> new MutableDocument("fromJSON", BaseDbTestKt.readJSONResource("array.json")));
     }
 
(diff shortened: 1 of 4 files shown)


common/test/java/com/couchbase/lite/DocumentTest.java
```diff
@@ -2712,6 +2712,16 @@ public void testMutableFromArray() {
...
     }
 
+    // Unsupported revision history
+    @Test
+    public void testGetRevisionHistory() throws CouchbaseLiteException {
+        MutableDocument mdoc = new MutableDocument("doc1");
+        Collection coll = getTestCollection();
+        coll.save(mdoc);
+        Document doc = coll.getDocument(mdoc.getId());
+        assertTrue(doc.getRevisionHistory().startsWith("1-"));
```
pasin, line 2722: I don't think this is correct for version vector. Probably just testing that it's not null and length > 0?

PR #352, opened 2024-10-24
CBL-6347: backport revHistory

bmeike:
Straightforward backport from 4.0

### DIFF
--- a/common/main/cpp/native_c4document.cc
+++ b/common/main/cpp/native_c4document.cc
@@ -153,6 +153,55 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Document_getSelectedRevID(JNI
 
 /*
  * Class:     com_couchbase_lite_internal_core_impl_NativeC4Document
+ * Method:    getRevisionHistory
+ * Signature: (JJJ[Ljava/lang/String;)Ljava/lang/String;
+ */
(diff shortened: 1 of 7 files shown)


common/main/cpp/native_c4document.cc
```diff
@@ -151,6 +151,58 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Document_getSelectedRevID(JNI
...
+        jlong maxRevs,
+        jobjectArray jBackToRevs) {
+    C4String *backToRevs = nullptr;
+    unsigned int nBackToRevs = 0;
+
+    // if jBackToRevs is non-null, a Java String[], convert it to a C array of C4Slice:
+    if (jBackToRevs != nullptr) {
+        jsize n = env->GetArrayLength(jBackToRevs);
+        if (env->EnsureLocalCapacity(std::min(n + 1, MaxLocalRefsToUse)) < 0)
+            return nullptr;
```
pasin, line 174: Not sure if this check is needed as I don't think we create any new references to java objects? I might misunderstood this function though.

pasin (approved):

common/main/cpp/native_c4document.cc
bmeike: You are absolutely right.  Removing it.

PR #354, opened 2024-11-15
CBL-6186: C4Replicator is a C4Peer

bmeike:
also: Don't print redundant stack trace on replicator status change


borrrden (approved):
Looks like this consists of mainly a bunch of finalizer removal, and cleaner queue additions that look quite similar to one another.  I don't know enough to say anything about the actual logic here but it seems sane.

pasin (approved):

PR #367, opened 2025-01-29
Partial Index

bmeike:
This is an almost direct cherry-pick of https://github.com/couchbase/couchbase-lite-java-common/pull/366
Just some stuff commented out and @Ignored until LiteCore catches up.

### DIFF
--- a/common/main/cpp/native_c4collection.cc
+++ b/common/main/cpp/native_c4collection.cc
@@ -254,7 +254,7 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Collection_getIndexesInfo(
 /*
  * Class:     com_couchbase_lite_internal_core_impl_NativeC4Collection
  * Method:    createValueIndex
- * Signature: (JLjava/lang/String;ILjava/lang/String;)V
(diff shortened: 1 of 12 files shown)


common/main/cpp/native_c4collection.cc
```diff
@@ -263,8 +263,13 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4Collection_createValueIndex(
         jlong coll,
         jstring jName,
         jint qLanguage,
-        jstring jqueryExpressions) {
+        jstring jqueryExpressions,
+        jstring where) {
     C4IndexOptions options = {};
+
+    jstringSlice whereSlice(env, where);
+    // options.where = whereSlice.c_str();
```
borrrden, line 271: Something I tend to do in this situation is put in a `#warning` line so that the compiler will say something like "where is not being applied".  Then the fact that this is not happening shows up in the compilation output.

borrrden (approved):

common/main/cpp/native_c4collection.cc
bmeike: That's a great idea.  I, instead, put a note in the tests that this has to be uncommented before the tests will work.

I didn't know about #warning, though.  I'll use it in the future.

PR #373, opened 2025-02-21
Tests for CBSE-19205

bmeike:
Two tests, one for each of the two failure modes.
Instrumentation to support them.
The same instrumentation supports the previously commented out NonASCII test.

... and, fwiw, I found another call to c4log_getDomain and protected it as well.  It is very unlikely that it would ever be called from within the callback.

### DIFF
--- a/common/test/java/com/couchbase/lite/logging/LogTest.kt
+++ b/common/test/java/com/couchbase/lite/logging/LogTest.kt
@@ -3,21 +3,19 @@ package com.couchbase.lite.logging
 import com.couchbase.lite.BaseDbTest
 import com.couchbase.lite.CBLError
 import com.couchbase.lite.CouchbaseLiteException
-import com.couchbase.lite.DataSource
 import com.couchbase.lite.Defaults
 import com.couchbase.lite.LogDomain
 import com.couchbase.lite.LogLevel
-import com.couchbase.lite.MutableDocument
-import com.couchbase.lite.QueryBuilder
-import com.couchbase.lite.SelectResult
+import com.couchbase.lite.getC4Db
+import com.couchbase.lite.internal.QueryLanguage
+import com.couchbase.lite.internal.core.C4Query
+import com.couchbase.lite.internal.core.C4TestUtils
 import com.couchbase.lite.internal.core.CBLVersion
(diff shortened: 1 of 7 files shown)


common/test/java/com/couchbase/lite/logging/LogTest.kt
```diff
@@ -547,33 +545,85 @@ class LogTest : BaseDbTest() {
...
+        val finishLatch = CountDownLatch(1)
+
+        val logSinks = assertNonNull(LogSinksImpl.getLogSinks())
+        logSinks.file = FileLogSink.Builder().setDirectory(scratchDirPath!!).build()
+
+        // If there is a file logger, an attempt to log a message from within the callback
+        // will call c4log_getDomain, which would deadlock the thread.
+        logSinks.c4Log.setCallbackInstrumentation { _: String?, _: Int, _: String? ->
+            startLatch.countDown()
+            Log.e(LogDomain.DATABASE, "BANG!!!")
```
pasin, line 562: Not sure if using Log.e() here will cause the dead lock in this test. It depends on whether Log.e() causes the log callback to be called or causes `c4log_getDomain()` to be called or not. I'm pointing this out because the platform's log() method will call to `ConsoleLogSink` and `CustomLogSink` directly without going through LiteCore.

If instead of calling Log.e(), call `C4TestUtils.forceLog()` or `new Database("foo")`, it will definitely hang.

pasin, line 562: Not sure if using Log.e() here will cause the dead lock in this test. It depends on whether Log.e() causes the log callback to be called or causes `c4log_getDomain()` to be called or not. I'm pointing this out because the platform's log() method will call to `ConsoleLogSink` and `CustomLogSink` directly without going through LiteCore.

If instead of calling Log.e(), call `C4TestUtils.forceLog()` or `new Database("foo")`, it will definitely hang.

bmeike: This is a test for the CBSE-19205 issue.  It is our analysis that lead us to suspect the call to `c4log_getDomain`.  What is failing for the customer is recursive logging.

pasin (approved):

PR #375, opened 2025-02-28
Better logging

bmeike:
- Better descriptions of the two new tests, now that I understand them.
- Comments on the LogSinks interface: esp the note about asynchronicity.
- Fixed testRecursiveLogging to check only for the reentrant case.
- Up the time limits on some of the Load tests so that they pass more reliably

### DIFF
--- a/common/test/java/com/couchbase/lite/logging/LogTest.kt
+++ b/common/test/java/com/couchbase/lite/logging/LogTest.kt
@@ -8,6 +8,7 @@ import com.couchbase.lite.LogDomain
 import com.couchbase.lite.LogLevel
 import com.couchbase.lite.getC4Db
 import com.couchbase.lite.internal.QueryLanguage
+import com.couchbase.lite.internal.core.C4Log
 import com.couchbase.lite.internal.core.C4Query
 import com.couchbase.lite.internal.core.C4TestUtils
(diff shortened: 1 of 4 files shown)


borrrden (approved):

common/test/java/com/couchbase/lite/logging/LogTest.kt
```diff
@@ -548,61 +549,102 @@ class LogTest : BaseDbTest() {
...
+            override fun onLogToCore(domain: LogDomain, level: LogLevel, message: String): Boolean {
+                success = false
+                return false // do not allow a recursive call!!  It will deadlock!
+            }
+
+            override fun onCallback(c4Domain: String?, c4Level: Int, message: String?): Boolean {
+                startLatch.countDown()
+
+                // Try to generate a recursive call to LiteCore logging
+                Log.e(LogDomain.DATABASE, "BANG!!!")
```
pasin, line 581: Probably this question is rephrasing what is described as the comment of this test but just want to make sure I understand it correctly.

Is it correct that this test tests that Log.e() will not cause `logToCore()` to be called as if that happens, it will cause a deadlock from `c4log_getDomain` call?

pasin (approved):

PR #379, opened 2025-03-11
Improved JVM attachement handling

bmeike:
All tests are passing... but this is C code, so I'd really appreciate a close read.

### DIFF
--- a/common/main/cpp/native_c4.cc
+++ b/common/main/cpp/native_c4.cc
@@ -63,10 +63,11 @@ bool litecore::jni::initC4Logging(JNIEnv *env) {
 }
 
 //-------------------------------------------------------------------------
-// Logging
+// Pre-initialization Logging
 //-------------------------------------------------------------------------
 
-// The default logging callback writes to stderr, or on Android to __android_log_write.
+// Before our logger is set up, or when something in the logging system fails,
+// log to stderr, or on Android to __android_log_write.
 // ??? Need to do something better for web service and Windows logging
 void vLogError(const char *fmt, va_list args) {
 #if defined(__ANDROID__)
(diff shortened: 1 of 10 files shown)


common/main/cpp/native_glue.cc
```diff
@@ -168,14 +171,37 @@ namespace litecore::jni {
...
 #ifdef JNI_VERSION_1_8
-        return gJVM->AttachCurrentThread(reinterpret_cast<void **>(env), NULL);
+                gJVM->AttachCurrentThread(reinterpret_cast<void **>(env), NULL);
 #else
-        return gJVM->AttachCurrentThread(env, nullptr);
+                gJVM->AttachCurrentThread(env, nullptr);
 #endif
+
+        if (ret == JNI_OK)
+            return JNI_EDETACHED;
```
pasin, line 194: This is confusing to me. Doesn't `JNI_EDETACHED` mean not attached? Why not returning `JNI_OK`?

common/main/cpp/native_c4.cc
```diff
@@ -98,43 +99,36 @@ void litecore::jni::jniLog(const char *fmt, ...) {
...
-        C4Warn("logCallback(): doRequestClose(): Failed to detach the current thread from a Java VM");
-    }
-    return false;
+
+//-------------------------------------------------------------------------
+// Logging Callback
+//-------------------------------------------------------------------------
+
+static bool detach(jint envState) {
+    return (envState == JNI_EDETACHED) && (JNI_OK == detachJVM("log"));
```
pasin, line 108: From the meaning of `JNI_EDETACHED`, I understand that JVM is not currently attached. If that true, should it be that `return (envState == JNI_EDETACHED) || (JNI_OK == detachJVM("log"));` instead?

common/main/cpp/native_glue.cc
borrrden: If I read this right, and this probably deserves a comment because of its unusualness, JNI_EDETACHED is being used here to indicate that the state was detached before the call to attach (possibly to do or not do cleanup later)

common/main/cpp/native_c4.cc
```diff
@@ -98,43 +99,36 @@ void litecore::jni::jniLog(const char *fmt, ...) {
...
-
-    if (getEnvStat == JNI_EDETACHED) {
-        if (attachCurrentThread(&env) != 0) {
-            jniLog("Logger: Failed to attach the current thread for logging");
-            return;
-        }
-    } else if (getEnvStat != JNI_OK) {
-        jniLog("Logger: Could not get the environment: %d", getEnvStat);
+    jint envState = attachJVM(&env, "logger");
+    if ((envState != JNI_OK) && (envState != JNI_EDETACHED))
```
borrrden, line 119: This is showing up in so many places now that it might even deserve a little inline method or macro, but it's not a big deal.

common/main/cpp/native_c4listener.cc
```diff
@@ -64,57 +64,34 @@ bool litecore::jni::initC4Listener(JNIEnv *env) {
...
-                    toJString(env, authHeader));
-            if (gJVM->DetachCurrentThread() != 0)
-                C4Warn("httpAuthCallback(): Failed to detach the current thread from a Java VM");
-        } else {
-            C4Warn("httpAuthCallback(): Failed to attaches the current thread to a Java VM");
-        }
+    jstring _header = toJString(env, authHeader);
+    jboolean ok = env->CallStaticBooleanMethod(cls_C4Listener, m_C4Listener_httpAuthCallback, (jlong) context, _header);
+    if (envState == JNI_EDETACHED) {
+        detachJVM("httpAuthCallback");
```
borrrden, line 74: You might consider passing the envState into this method to save on the repeated ifs throughout these files, but that's also not a big deal.

common/main/cpp/native_glue.cc
pasin: I feel that the return here is inconsistent with the return in line 179. Basically if it's already attached when calling GetEnv, it returns `JNI_OK` in line 179 but it returns `JNI_EDETACHED` in line 194 if the attach result is successful.

common/main/cpp/native_c4.cc
pasin: When comparing this with the old code, look like the behavior is the same BUT it's just confusing as when I read the code I thought that `envState` is the current state but it turns out that it tells you what to do.

common/main/cpp/native_glue.cc
pasin: I kind of get it now. Agree with Jim that it deserves comment as it's an unusual return. I understand that the returned status is the result of the operation but what it returns now is either the previous state or the action that should be done next.

bmeike: Jim is correct.  I will add a comment.

common/main/cpp/native_c4.cc
bmeike: See above: JNI_EDETACHED describes the state of the thread when it entered the callback, not its current state.  If it was JNI_EDETACHED, it needs to be detached again, on the way out.

borrrden (approved):

pasin (approved):

PR #382, opened 2025-03-26
CBL-6879: Link with 16KB page size

bmeike:
I don't know of a way to demonstrate that this works....
... and I'm a bit concerned because Jim's version uses a second "-Wl" in front of the second arg.  My build failed when I tried to do that ( ld: error: unknown argument '-Wl' )
I am assuming that the first -Wl on the line directs the rest of the line to the linker, which does not, of course, understand "-Wl"

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -83,7 +83,7 @@ target_include_directories( LiteCoreJNI PUBLIC
 
 set_target_properties(
     LiteCoreJNI PROPERTIES LINK_FLAGS
-    "-Wl,--version-script=${PROJECT_SOURCE_DIR}/jni_android.ld"
+    "-Wl,-z,max-page-size=16384,--version-script=${PROJECT_SOURCE_DIR}/jni_android.ld"
 )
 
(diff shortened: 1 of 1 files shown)


bmeike:
I have verified that the JNI is linked on 16kb boundaries: ```
 >/Users/blakemeike/Library/Android//sdk/ndk/29.0.13113456/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-objdump -p ./lib/build/intermediates/cxx/Debug/2i481c2x/obj/arm64-v8a/libLiteCoreJNI.so | grep LOAD
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**14
    LOAD off    0x0000000000040e40 vaddr 0x0000000000044e40 paddr 0x0000000000044e40 align 2**14
    LOAD off    0x0000000000041b18 vaddr 0x0000000000049b18 paddr 0x0000000000049b18 align 2**14
12:20>
```

borrrden (approved):

PR #384, opened 2025-04-16
Release candidate 3.2.3 rc 2

bmeike:
Annotate Document.getTimestamp as @Volatile
Quiet some C4Peer hysterics
Tests that depend on the extension package are skipped if it cannot be loaded…

The actually changes here are very minor.


borrrden (commented):
I don't actually see the part that marks getTimestamp as volatile.

bmeike:
I left the annotation out of the commit.  Just force pushed it.

pasin (approved):

PR #390, opened 2025-05-12
CBL-6978: Cancel pending conflict resolutions, on Replicator.stop

bmeike:


pasin (approved):
Before removing itself from the database's active replicators, does it also wait until all of the conflict resolutions are done / cancelled? I understand that this logic is not part of this PR but the existing logic. Just want to check how it is currently handled.

bmeike:
As Pasin points out there is a race when the replicator stops, the CRs are still running, and the client code tries to close the DB.  I've filed [BL-7001](https://jira.issues.couchbase.com/browse/CBL-7001)

PR #391, opened 2025-05-30
Release 3.2.4

bmeike:
The tests are a gigantic change. There is, however, only one change to production code, in the file AbstractReplicator.java.  It just doesn't get the connection options unless they are needed.


pasin (approved):
I only reviewed the change in the AbstractReplicator.

PR #398, opened 2025-06-25
CBL-7081: Multipeer Auth

bmeike:
I think this is code complete.  Everything else will just be debugging...  and, setting expectations, ther will undoubtedly be plenty of that...

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/core/C4Log.java
+++ b/common/main/java/com/couchbase/lite/internal/core/C4Log.java
@@ -118,10 +118,10 @@ public final class C4Log {
         m.put(C4Constants.LogDomain.BLOB, LogDomain.DATABASE);
         m.put(C4Constants.LogDomain.CHANGES, LogDomain.DATABASE);
         m.put(C4Constants.LogDomain.DATABASE, LogDomain.DATABASE);
-        m.put(C4Constants.LogDomain.DISCOVERY, LogDomain.NETWORK);
+        m.put(C4Constants.LogDomain.DISCOVERY, LogDomain.PEER_DISCOVERY);
         m.put(C4Constants.LogDomain.ENUM, LogDomain.QUERY);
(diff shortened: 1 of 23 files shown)


common/main/java/com/couchbase/lite/internal/core/C4Log.java
```diff
@@ -167,6 +167,8 @@ public boolean isInUse() {
         m.put(LogDomain.REPLICATOR, C4Constants.LogDomain.SYNC);
         m.put(LogDomain.QUERY, C4Constants.LogDomain.QUERY);
         m.put(LogDomain.LISTENER, C4Constants.LogDomain.LISTENER);
+        m.put(LogDomain.MULTIPEER, C4Constants.LogDomain.DISCOVERY);
+        m.put(LogDomain.PEER_DISCOVERY, C4Constants.LogDomain.P2P);
```
borrrden, line 171: Is this supposed to be a reverse mapping of the above?  If so it doesn't seem consistent:

Discovery -> Peer Discovery but Peer Discovery -> P2P for example

borrrden (approved):

common/test/java/com/couchbase/lite/mock/SocketMocks.kt
```diff
@@ -43,6 +43,7 @@ open class MockCore : SocketToCore {
     private val mutex = Object()
     override fun getLock() = mutex
     override fun init(listener: SocketFromCore): Unit = TODO("Not yet implemented")
+    override fun gotPeerCertificate(certData: ByteArray, hostname: String): Boolean = TODO("Not yet implemented")
```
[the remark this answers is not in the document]
bmeike: no

common/main/java/com/couchbase/lite/internal/core/C4Log.java
bmeike: Yes.  Good catch.

PR #418, opened 2025-09-01
CBL-7381 : Remove DatabaseConfiguration's mmapEnabled property

Aniket392:
- Removed API for setting MmapEnabled
- Updated testcases for checking default MmapEnabled

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractDatabaseConfiguration.java
+++ b/common/main/java/com/couchbase/lite/AbstractDatabaseConfiguration.java
@@ -30,32 +30,29 @@ abstract class AbstractDatabaseConfiguration {
     //---------------------------------------------
     private String dbDirectory;
     private boolean fullSync;
-    private boolean mmapEnabled;
 
     //---------------------------------------------
     // Constructors
     //---------------------------------------------
     protected AbstractDatabaseConfiguration() {
-        this(null, Defaults.Database.FULL_SYNC, Defaults.Database.MMAP_ENABLED);
+        this(null, Defaults.Database.FULL_SYNC);
     }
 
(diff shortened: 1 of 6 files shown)


common/main/java/com/couchbase/lite/AbstractDatabaseConfiguration.java
```diff
@@ -114,17 +114,6 @@ public DatabaseConfiguration setFullSync(boolean isFullSync) {
...
-     * Nmory-mapped database files are, currently, enabled by default, except on MacOS
-     * where they <b>cannot</b> be enabled at all.
-     */
-    @NonNull
-    public DatabaseConfiguration setMMapEnabled(boolean mmapEnabled) {
-        this.mmapEnabled = mmapEnabled;
-        return getDatabaseConfiguration();
-    }
-
     public boolean isMMapEnabled() { return mmapEnabled; }
```
borrrden, line 117: Why is this not also deleted?

borrrden (changes requested):

common/main/java/com/couchbase/lite/AbstractDatabaseConfiguration.java
Aniket392: Currently Mmapenabled can be changed on the basis of default MMAP_ENABLED and apis to user
Just removed the apis to user and set the MMAP_ENABLED to false. This helps in keeping the structure same.
So thats why we do not need to remove the isMmapEnabled, which is used in various dependent api.

pasin (approved):

common/main/java/com/couchbase/lite/AbstractDatabaseConfiguration.java
Aniket392: Updated as per comment

PR #421, opened 2025-09-04
CBL-7412 : Couchbase Lite uses discontinued convenience method from okhttp

Aniket392:
- Removed the import and usage of check duration from internal api of okhttp
- Used method for same checkDuration Logic

### DIFF
--- a/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
+++ b/common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
@@ -30,7 +30,6 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import okhttp3.internal.Util;
 
 import com.couchbase.lite.internal.BaseReplicatorConfiguration;
 import com.couchbase.lite.internal.ImmutableReplicatorConfiguration;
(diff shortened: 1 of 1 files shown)


common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
```diff
@@ -71,11 +70,19 @@ public enum ReplicatorType {PUSH_AND_PULL, PUSH, PULL}
...
         return heartbeat;
     }
 
+    private static void checkDuration(String name, long duration) {
+        Preconditions.assertNotNegative(duration, name);
+        final long millis = TimeUnit.SECONDS.toMillis(duration);
+        if (millis > Integer.MAX_VALUE || (millis == 0L && duration > 0L)) {
+            throw new IllegalArgumentException(name + (millis > Integer.MAX_VALUE ? " too large." : " too small."));
+        }
+    }
```
pasin, line 84: If this is logic is used by the other place in the future, we could move this check to Preconditions class.

pasin (approved):

common/main/java/com/couchbase/lite/AbstractReplicatorConfiguration.java
pasin: Can you also check if there are any other places that use `okhttp3.internal.Util`? In general, we don't want to use any internal API from 3rd parties.

Aniket392: > If this is logic is used by the other place in the future, we could move this check to Preconditions class.

Yes, moving it into preconditions makes sense, will update the PR

PR #430, opened 2025-09-18
CBL-7453 : Fix URLEndpointListenerTest failure

pasin:
* Fix crash when generating a self-signed certificate without an issuer. In native_c4listener::generateCertificate(), do not release issuerKey if it is the same C4KeyPair passed from Java (owned and managed by the Java side).

* Fix acceptOnlySelfSignedServerCert not working due to incorrect parameter order passed to the CBLTrustManager constructor.


pasin:
We need to port the fix to 3.3 as well. Not sure why android tests are passed.

borrrden (approved):

PR #431, opened 2025-09-19
CBL-7454 : URLEndpointListenerTests Failed on Java

Aniket392:
-  Fix crash when generating a self-signed certificate without an issuer. In native_c4listener::generateCertificate(), do not release issuerKey if it is the same C4KeyPair passed from Java (owned and managed by the Java side).

- Fix acceptOnlySelfSignedServerCert not working due to incorrect parameter order passed to the CBLTrustManager constructor.


pasin:
Will need one for android/release/3.3 as well.

PR #433, opened 2025-10-04
CBL-7493 : Fix wrong argument order when creating CBLTrustManager

pasin:
* In AbstractCBLWebSocket, swap the order between `acceptAllCerts` and `acceptOnlySelfSignedServerCert`.

* Update the order in Java’s CBLTrustManager to be in sync with the Android version (inline with the order in AbstractCBLTrustManager).


pasin:
This is a simple fix and I have verified that the failed tests are passed. I'm merging this myself.

PR #456, opened 2026-02-23
P2P Bluetooth peer discovery changes

Aniket392:
- Adding c4Peerdiscovery implementation, implementing all methods and connecting it to java
- BT connections discovery logic from both native as well as java side
- Writing helper class like MetadataHelper, to convert java map object
- Adding native implementation to BluetoothPeer
- Updating native_c4multipeerreplicator to take protocol parameter

### DIFF
--- a/android/main/AndroidManifest.xml
+++ b/android/main/AndroidManifest.xml
@@ -20,6 +20,7 @@ limitations under the License.
     xmlns:tools="http://schemas.android.com/tools"
     >
 
+    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
--- a/common/main/cpp/native_c4multipeerreplicator.cc
+++ b/common/main/cpp/native_c4multipeerreplicator.cc
@@ -101,7 +101,7 @@ namespace litecore::jni {
             m_C4MultipeerReplicator_createPeerInfo = env->GetStaticMethodID(
                     cls_C4MultipeerReplicator,
                     "createPeerInfo",
-                    "([B[BZ[[BLcom/couchbase/lite/internal/core/C4ReplicatorStatus;)Lcom/couchbase/lite/PeerInfo;");
+                    "([B[BZ[[BLjava/util/Set;Lcom/couchbase/lite/internal/core/C4ReplicatorStatus;)Lcom/couchbase/lite/PeerInfo;");
 
             if (m_C4MultipeerReplicator_createPeerInfo == nullptr)
                 return false;
@@ -125,7 +125,7 @@ namespace litecore::jni {
             m_C4MultipeerReplicator_onPeerDiscovered = env->GetStaticMethodID(
                     cls_C4MultipeerReplicator,
                     "onPeerDiscovered",
-                    "(J[BZ)V");
+                    "(J[BLcom/couchbase/lite/MultipeerTransport;Z)V");
 
             if (m_C4MultipeerReplicator_onPeerDiscovered == nullptr)
                 return false;
@@ -133,7 +133,7 @@ namespace litecore::jni {
             m_C4MultipeerReplicator_onReplicatorStatusChanged = env->GetStaticMethodID(
                     cls_C4MultipeerReplicator,
                     "onReplicatorStatusChanged",
-                    "(J[BZLcom/couchbase/lite/internal/core/C4ReplicatorStatus;)V");
+                    "(J[BLcom/couchbase/lite/MultipeerTransport;ZLcom/couchbase/lite/internal/core/C4ReplicatorStatus;)V");
 
             if (m_C4MultipeerReplicator_onReplicatorStatusChanged == nullptr)
                 return false;
@@ -141,7 +141,7 @@ namespace litecore::jni {
             m_C4MultipeerReplicator_onDocumentEnded = env->GetStaticMethodID(
                     cls_C4MultipeerReplicator,
                     "onDocumentEnded",
-                    "(J[BZ[Lcom/couchbase/lite/internal/core/C4DocumentEnded;)V");
+                    "(J[BLcom/couchbase/lite/MultipeerTransport;Z[Lcom/couchbase/lite/internal/core/C4DocumentEnded;)V");
 
             if (m_C4MultipeerReplicator_onDocumentEnded == nullptr)
                 return false;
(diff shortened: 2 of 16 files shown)


android/main/AndroidManifest.xml
```diff
@@ -20,6 +20,7 @@ limitations under the License.
     xmlns:tools="http://schemas.android.com/tools"
     >
 
+    <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
```
pasin: What does this mean to users who don't use MultipeerReplicator with Bluetooth?

common/main/cpp/native_c4multipeerreplicator.cc
```diff
@@ -525,9 +528,7 @@ JNICALL Java_com_couchbase_lite_internal_core_impl_NativeC4MultipeerReplicator_c
     params.peerGroupID = groupId;
 
     // Protocols:
-    C4String protocols[] = {kPeerSyncProtocol_DNS_SD};
-    params.protocols = protocols;
-    params.protocolsCount = sizeof(protocols) / sizeof(protocols[0]);
+    params.protocols = kPeerSyncProtocol_BluetoothLE;
```
pasin: Need to be passed as an array of transport values.

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+                stopPublishing();
+            }
+        }
+
+        void addDiscoveredPeer(C4Peer* peer, bool moreComing = false) {
+            addPeer(peer, moreComing);
+        }
+
+        void removeDiscoveredPeer(std::string id, bool moreComing = false) {
+            removePeer(id, moreComing);
```
pasin: Cannot use `removePeer(id, moreComing)` directly?

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+        const char* peerIdChars = env->GetStringUTFChars(peerId, nullptr);
+        if (peerIdChars) {
+            peerIdStr = std::string(peerIdChars);
+            env->ReleaseStringUTFChars(peerId, peerIdChars);
+        }
+    }
+
+    // === CALL LIBRARY API ===
+    C4PeerDiscovery& discovery = provider->discovery();
+    fleece::Retained<C4Peer> peer = discovery.peerWithID(peerIdStr);
```
pasin: This peer is a Retained object and its ref count will be decreased by 1 after this function is returned (done). In line 456, its raw pointer is being passed to the Java object and as I understand, the Java object will keep the raw pointer to use later. 

The problem with this is that the peer's raw pointer could be freed any time depending on the logic inside the discovery. It's unsafe to use by the Java object that keeps the raw pointer and the crash is due to happen. 

Therefore, it's better to keep the raw pointer with ref count + 1 by calling peer.detach() to get that. Then pass the pointer with ref count + 1 to the Java object instead. However, when Java object is done, it will need to call to release the pointer as well. 

To implement that, you can add a C function such as C4Peer_Release(C4Peer* peer) for Java to called via the JNI. Then the Java class needs to implement C4Peer if I understand. I'm not sure about the Cleaner stuff but I think you can follow some of the existing classes that do that.

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+            env->ReleaseByteArrayElements(peerId, bytes, JNI_ABORT);
+        }
+    }
+
+    // Get the C4Socket from the pointer
+    C4Socket* socket = (C4Socket*)socketPtr;
+    if (!socket) return;
+
+    // Find the peer in the discovery system
+    auto peer = provider->discovery().peerWithID("peerIdObj");
```
pasin: Is `peerWithID("peerIdObj")` correct with the "peerIdObject" string?

pasin (changes requested):

pasin (changes requested):

common/main/cpp/native_c4multipeerreplicator.cc
```diff
@@ -688,7 +689,7 @@ JNICALL Java_com_couchbase_lite_internal_core_impl_NativeC4MultipeerReplicator_g
             m_C4MultipeerReplicator_createPeerInfo,
             jpeerId,
             certChain,
-            info->online ? JNI_TRUE : JNI_FALSE,
+            JNI_TRUE,
```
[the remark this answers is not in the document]
Aniket392: Updated as shared

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+            if ((envState != JNI_OK) && (envState != JNI_EDETACHED))
+                return;
+
+            // Call Java BLE service to start scanning
+            env->CallStaticVoidMethod(
+                    cls_C4PeerDiscoveryProvider,
+                    m_C4PeerDiscoveryProvider_startBrowsing,
+                    _contextToken);
+
+
```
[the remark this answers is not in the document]
Aniket392: Removed extra spaces

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+                    m_C4PeerDiscoveryProvider_stopPublishing,
+                    _contextToken);
+
+            if (envState == JNI_EDETACHED) {
+                detachJVM("stopPublishing");
+            }
+        }
+
+
+
```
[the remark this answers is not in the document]
Aniket392: Removed extra lines

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -0,0 +1,583 @@
...
+    // Get the C4Socket from the pointer
+    C4Socket* socket = (C4Socket*)socketPtr;
+    if (!socket) return;
+
+    // Find the peer in the discovery system
+    auto peer = provider->discovery().peerWithID("peerIdObj");
+    if (peer) {
+        // Notify about the incoming connection
+        bool accepted = provider->notifyIncomConnection(peer.get(), socket);
```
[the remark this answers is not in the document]
Aniket392: Currently removed this
Will update with Socketfactory

pasin (changes requested):

pasin (changes requested):

pasin (approved):

PR #457, opened 2026-03-11
BT Socket Factory creation and connecting with java code

Aniket392:
- Add C4BTSocketFactory, registers C4SocketFactory with LiteCore and handles all JNI callbacks (open, write, close, requestClose, dispose, attached)
- Add LiteCoreBTSocket, bridges BluetoothSocket L2CAP channel with LiteCore's C4Socket; implements SocketFromCore with read loop, write queue, and flow control
- Add NativeC4BTSocketFactory, JNI implementation for nRegisterFactory and nFromNative


pasin (commented):
There are lot of detail and setup in this PR, and I think we will need to setup an person review. 

However, I have looked up where we put the current WebSocketFactory and CBLWebSocket classes. I think we should move LiteCoreBTSocket and C4BTSocketFactory to the **EE** repo under android's com.couchbase.lite.internal.<bluetooth packages where the other classes are at>. We may need to rename it without C4 and Core prefix (e.g. BluetoothSocketFactory and BluetoothSocket : Use the names that are consistent with others). 

As a reference, where CBLWebSocket / SocketFactory are :

https://github.com/couchbase/couchbase-lite-java-common/blob/b7bcaf74dd4f4752ba75e05f5adaaf6988931df6/android/main/java/com/couchbase/lite/internal/replicator/CBLWebSocket.java 

https://github.com/couchbaselabs/couchbase-lite-java-ee-root/blob/master/ee/common/main/java/com/couchbase/lite/internal/SocketFactory.java

common/main/cpp/native_c4btsocketfactory.cc
```diff
@@ -0,0 +1,296 @@
...
+    addr.port     = (uint16_t) jport;
+    addr.path     = path;
+
+    void* context = (void*) jtoken;
+    C4SocketFactory factory = kBTSocketFactory;
+    factory.framing = (C4SocketFraming) jframing;
+    factory.context = context;
+
+    C4Socket* c4socket = c4socket_fromNative(factory, context, &addr);
+    c4socket_retain(c4socket);
```
pasin: The socket needs to be wrapped with TLS enabled socket to get the TLS encryption. This is what is done in iOS but currently c4SocketFromNativeWithTLS () is in the TLSCodec.hh, and the header is not included in the package as a public header.

```
Ref<C4Socket> C4PeerDiscoveryProvider::createIncomingSocketWithTLS(C4SocketFactory const& factory,
                                                                   void* nativeHandle,
                                                                   C4Address const& address)
{
    return net::c4SocketFromNativeWithTLS(factory, nativeHandle, address, _discovery.tlsContext().get());
}
```

common/main/cpp/native_c4peerdiscoveryprovider.cc
```diff
@@ -498,6 +513,71 @@ Java_com_couchbase_lite_internal_core_impl_NativeC4PeerDiscoveryProvider_peersWi
...
+    };
+
+    std::optional<C4SocketFactory> factory = provider->getSocketFactory();
+    if (!factory.has_value()) {
+        C4Warn("nCreateIncomingSocket: no socket factory");
+        delete ctx;
+        return;
+    }
+
+    fleece::Ref<C4Socket> socket = provider->createIncomingSockWithTLS(
```
pasin: I'm not sure I understand the purpose of this `Java_com_couchbase_lite_internal_core_impl_NativeC4PeerDiscoveryProvider_createIncomingSocket` function as it doesn't return the created socket. Can you explain when this function is called?

pasin (approved):

common/main/cpp/native_c4peerdiscoveryprovider.cc
pasin: I have a chat with Aniket about this. The flow is to create it and give it to LiteCore via a notifyXXX() function and LiteCore will maintain the lifetime of the socket.

borrrden:
As you may tell, this went through weeks of exercise and rewriting to get into the state it is now, so consider it reviewed.

PR #474, opened 2026-05-07
Update digest algorithm and fix predictive query tests

pasin:
Updated digest algorithm and fixed predictive query tests after updating LiteCore to 3.2.6-1.


pasin:
Updated digest algorithm and fixed predictive query tests after updating LiteCore to 3.2.6-1.

PR #475, opened 2026-05-08
CBL-8389 : Add Kotlin serialization-based APIs

snej:
Implements Kotlin extension methods that use serialization to allow developers to work with documents and query rows as native Kotlin classes. (Equivalent to the [Swift API extensions](https://docs.google.com/document/d/1SDfFpKg6mFhkfQZrez2Cbz57QgNjPNcGSR9DYdJM4PQ/edit?tab=t.0#heading=h.9bvjqm19w1mm) implemented last year.)

## Query Results & ResultSets

```kotlin
/** Uses Kotlin Serialization to create an object of type [T] from the query row.
 *  If the [key] parameter is non-null, it uses the row's value for that key as the source,
 *  instead of the entire row.
 *
 *  For example, if your query is `SELECT name, age, shoeSize FROM people` and you have a
 *  serializable Person class with properties name, age and shoeSize, you can call:
 *  `val person = result.data<Person>()`
 *
 *  Or you could use the query `SELECT * as person FROM people` and create the Person with
 *  `val person = result.data<Person>("person")`.
 */
@ExperimentalSerializationApi
inline fun <reified T> Result.data(key: String? = null): T

/** Returns the query rows transformed by Kotlin Serialization into instances of [T]. */
@ExperimentalSerializationApi
inline fun <reified T> ResultSet.data(key: String? = null): Sequence<T>
```

## Documents

Document model classes must implement a new interface `DocumentModel` that defines a `documentMeta` property that stores the document id and revID, which is needed in order to save the document.

```kotlin
/** Document model classes must implement this interface.
 *  It adds a [documentMeta] property that's used by Couchbase Lite. */
interface DocumentModel {
    /** This tags the model instance with the document ID and revision it was read from,
     *  which enables conflict detection when it's later saved.
     *  You may read this property, but DO NOT alter it.
     *  It should be implemented as a stored property defaulting to `null`, for example:
     *  `@Transient override var documentMeta: DocumentMeta? = null` */
    @Transient var documentMeta: DocumentMeta?
}

/** Stores the Couchbase Lite metadata of a document. Used by the [DocumentModel] interface. */
class DocumentMeta internal constructor(val collection: Collection?,
                                        val id: String,
                                        val revisionID: String)
```

Documents can be loaded and saved as model instances:

```kotlin
/** Gets an existing document with the given ID, and uses Kotlin Serialization to create an
 *  instance of class [T] from it. [T] must implement [DocumentModel].
 *  If a document with the given ID doesn't exist in the collection, returns null. */
@ExperimentalSerializationApi
inline fun <reified T: DocumentModel> Collection.getDocumentAs(id: String): T? =
    getDocumentAs(id, serializer())

/** Saves a [DocumentModel] instance as a document in the collection, with a specified conflict handler.
 *  If the model's [DocumentModel.documentMeta] property is null, it will be saved as a new document with the
 *  given [docID], which must not be null.
 *  Otherwise the [DocumentModel.documentMeta] property determines the document ID and prior revision ID, and the
 *  [docID] parameter should be null.
 *  After a successful save, the [DocumentModel.documentMeta] property is updated to the current state. */
@ExperimentalSerializationApi
inline fun <reified T: DocumentModel> Collection.save(model: T,
                                                      docID: String? = null,
                                                      noinline conflictHandler: ModelConflictHandler<T>? = null) =
```

## Lower-Level

Both of these use some new internal functions that serialize Kotlin classes to and from Fleece.

```kotlin
/** Uses Kotlin Serialization to encode an arbitrary object or collection to Fleece.
 *  @param value  The object to encode.
 *  @param encoder  A Fleece [FLEncoder] to use; defaults to a fresh instance.
 *  @return  The encoded Fleece data. */
@ExperimentalSerializationApi
inline fun <reified T> serializeToFleece(value: T, encoder: FLEncoder? = null): ByteArray

/** Decodes an object from a Fleece [FLValue] using Kotlin Serialization. */
@ExperimentalSerializationApi
inline fun <reified T> deserializeFromFleece(root: FLValue): T
```

### DIFF
--- a/common/main/java/com/couchbase/lite/Document.java
+++ b/common/main/java/com/couchbase/lite/Document.java
@@ -622,6 +622,13 @@ public class Document implements DictionaryInterface, JSONEncodable, Iterable<St
         }
     }
 
+    // for use by CollectionExtensions.kt
+    void setContent(@NonNull FLDict data, boolean mutable) {
+        synchronized (lock) {
+            setContentLocked(data, mutable);
+        }
+    }
+
     @GuardedBy("lock")
     private void updateC4DocumentLocked(@Nullable C4Document c4Doc) {
         if (c4Document == c4Doc) { return; }
--- a/common/main/java/com/couchbase/lite/Document.java
+++ b/common/main/java/com/couchbase/lite/Document.java
@@ -33,6 +33,7 @@ import com.couchbase.lite.internal.core.C4Document;
 import com.couchbase.lite.internal.fleece.FLDict;
 import com.couchbase.lite.internal.fleece.FLEncoder;
 import com.couchbase.lite.internal.fleece.FLSliceResult;
+import com.couchbase.lite.internal.fleece.FLValue;
 import com.couchbase.lite.internal.fleece.JSONEncodable;
 import com.couchbase.lite.internal.fleece.MRoot;
 import com.couchbase.lite.internal.utils.ClassUtils;
@@ -132,6 +133,12 @@ public class Document implements DictionaryInterface, JSONEncodable, Iterable<St
     @GuardedBy("lock")
     @Nullable
     private String revId;
+
+    // Set by setData(FLSliceResult,boolean) to keep the Fleece backing store from being GC'd.
+    // (This is kind of a hack, and it's only used ephemerally by Kotlin serialization.)
+    @GuardedBy("lock")
+    @Nullable
+    private FLSliceResult extraBackingStore;
 
     //---------------------------------------------
     // Constructors
@@ -623,8 +630,10 @@ public class Document implements DictionaryInterface, JSONEncodable, Iterable<St
     }
 
     // for use by CollectionExtensions.kt
-    void setContent(@NonNull FLDict data, boolean mutable) {
+    void setContent(@NonNull FLSliceResult fleeceData, boolean mutable) {
         synchronized (lock) {
+            var data = FLValue.fromData(fleeceData).asFLDict();
+            extraBackingStore = fleeceData;
             setContentLocked(data, mutable);
         }
     }
--- a/common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
+++ b/common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
@@ -54,7 +54,7 @@ fun <T: DocumentModel> Collection.getDocumentAs(id: String, deserializer: Deseri
 
 /** Saves a [DocumentModel] instance as a document in the collection, with a specified conflict handler.
  *  If the model's [DocumentModel.documentMeta] property is null, it will be saved as a new document with the
- *  given [docID], which must not be null.
+ *  given [docID], or if [docID] is null, with an auto-generated ID.
  *  Otherwise the [DocumentModel.documentMeta] property determines the document ID and prior revision ID, and the
  *  [docID] parameter should be null.
  *  After a successful save, the [DocumentModel.documentMeta] property is updated to the current state. */
@@ -75,7 +75,6 @@ fun <T: DocumentModel> Collection.save(model: T,
     val meta = model.documentMeta
     val doc: MutableDocument
     if (meta == null) {
-        require(docID != null) { "docID argument must be given when saving a new document" }
         doc = MutableDocument(docID)
     } else {
         require(meta.collection == this || meta.collection == null) {"saving document to wrong collection"}
--- a/common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
+++ b/common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
@@ -1,0 +1,153 @@
+//
+// Copyright (c) 2026 Couchbase, Inc All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+@file:OptIn(ExperimentalSerializationApi::class)
+
+package com.couchbase.lite
+
+import com.couchbase.lite.internal.core.C4Document
+import com.couchbase.lite.internal.fleece.*
+import kotlinx.serialization.*
+
+
+/** Document model classes must implement this interface.
+ *  It adds a [documentMeta] property that's used by Couchbase Lite. */
+interface DocumentModel {
+    /** This tags the model instance with the document ID and revision it was read from,
+     *  which enables conflict detection when it's later saved.
+     *  You may read this property, but DO NOT alter it.
+     *  It should be implemented as a stored property defaulting to `null`, for example:
+     *  `@Transient override var documentMeta: DocumentMeta? = null` */
+    @Transient var documentMeta: DocumentMeta?
+}
+
+/** Stores the Couchbase Lite metadata of a document. Used by the [DocumentModel] interface. */
+class DocumentMeta internal constructor(val collection: Collection?, // [Result] leaves it null
+                                        val id: String,
+                                        val revisionID: String)
+
+
+/** Gets an existing document with the given ID, and uses Kotlin Serialization to create an
+ *  instance of class [T] from it. [T] must implement [DocumentModel].
+ *  If a document with the given ID doesn't exist in the collection, returns null. */
+@ExperimentalSerializationApi
+inline fun <reified T: DocumentModel> Collection.getDocumentAs(id: String): T? =
+    getDocumentAs(id, serializer())
+
+@ExperimentalSerializationApi
+fun <T: DocumentModel> Collection.getDocumentAs(id: String, deserializer: DeserializationStrategy<T>): T? =
+    modelFromC4Doc(this, id, getC4Document(id), deserializer)
+
+
+/**
+ * Saves a [DocumentModel] instance as a document in the collection, with a specified conflict handler.
+ * If the model's [DocumentModel.documentMeta] property is null, it will be saved as a new document with the
+ * given [docID], which must not be null.
+ * Otherwise the [DocumentModel.documentMeta] property determines the document ID and prior revision ID, and the
+ * [docID] parameter should be null.
+ * After a successful save, the [DocumentModel.documentMeta] property is updated to the current state. */
+@ExperimentalSerializationApi
+inline fun <reified T: DocumentModel> Collection.save(model: T,
+                                                      docID: String? = null,
+                                                      noinline conflictHandler: ModelConflictHandler<T>? = null) =
+    save(model, serializer(), serializer(), docID, conflictHandler)
+
+@ExperimentalSerializationApi
+fun <T: DocumentModel> Collection.save(model: T,
+                                       serializer: SerializationStrategy<T>,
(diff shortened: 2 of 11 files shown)


snej:
I just noticed that Swift has model-based `Collection.delete` and `Collection.purge` methods; I'll add those.

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
```diff
@@ -0,0 +1,177 @@
...
+    getDocumentAs(id, serializer())
+
+@ExperimentalSerializationApi
+fun <T: DocumentModel> Collection.getDocumentAs(id: String, deserializer: DeserializationStrategy<T>): T? =
+    modelFromC4Doc(this, id, getC4Document(id), deserializer)
+
+
+/** Saves a [DocumentModel] instance as a document in the collection, with a specified conflict handler.
+ *  If the model's [DocumentModel.documentMeta] property is null, it will be saved as a new document with the
+ *  given [docID], which must not be null.
```
pasin: In Swift, we also support id == nil as an auto generated id for a newly created doc. Can we allow that here as well?

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
```diff
@@ -0,0 +1,177 @@
...
+                                       deserializer: DeserializationStrategy<T>,
+                                       docID: String? = null,
+                                       conflictHandler: ModelConflictHandler<T>? = null): Boolean
+{
+    // Get or create the Document:
+    val meta = model.documentMeta
+    val doc: MutableDocument
+    if (meta == null) {
+        require(docID != null) { "docID argument must be given when saving a new document" }
+        doc = MutableDocument(docID)
```
pasin, line 85: Can we also allow null docID here (an auto id (uuid) will be generated)?

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
```diff
@@ -0,0 +1,177 @@
...
+    model.documentMeta = DocumentMeta(collection, docID, c4doc.revID!!)
+    return model
+}
+
+
+/** Extension of [MutableDocument], that updates its content from a [DocumentModel] object. */
+private fun <T:DocumentModel> MutableDocument.setContentFromModel(model: T, serializer: SerializationStrategy<T>) {
+    val body = serializeToFleece(serializer, model)
+    val root = FLValue.fromData(body).asFLDict()
+    setContent(root, false)
```
pasin: Does the body need to be retained as well? 

I read iOS code, and iOS is doing something similar. If I read the code correctly, Swift's DocumentEncoder calls FLEncoder_FinishDoc that returns an FLDoc, get the dict from the FLDoc, set the dict to the CBL doc, and abandon the FLDoc. So maybe the data is retained by the dict?

```
- (BOOL)finishIntoDocument:(CBLDocument*)document error:(NSError**)outError {
    FLError error {};
    FLDoc fldoc = FLEncoder_FinishDoc(_encoder, &error);
    if (!fldoc) {
        return convertError(error, outError);
    }
    Doc doc { fldoc };
    Dict fleeceData = doc.asDict();
    if (!fleeceData) {
        return NO;
    }
    [document setFleece: (FLDict)fleeceData];
    return YES;
}
```

snej: That Obj-C code above leaks the FLDoc, unfortunately. (The `Doc` constructor retains the FLDoc, then releases it when the function returns. But nothing releases the +1 ref returned by `FLEncoder_FinishDoc`.

That leak is the only reason it doesn't crash, I think, unless the `setFleece:` method is doing something clever like explicitly retaining the FLDict.

snej: I confess I can't figure out how the memory management of Fleece objects works in the Java code. Do you or Jim have any better knowledge of it?

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
snej: Done!

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
snej: Done!

common/main/kotlin/com/couchbase/lite/internal/fleece/FLValueExtensions.kt
```diff
@@ -0,0 +1,107 @@
...
+    override val size: Int = dict.count().toInt()
+
+    override fun isEmpty(): Boolean = (dict.count() == 0L)
+
+    override fun get(key: String): FLValue? = dict.get(key)
+
+    override fun containsKey(key: String): Boolean = (dict.get(key) != null)
+
+    override fun containsValue(value: FLValue): Boolean {
+        val iter = dict.iterator()
```
borrrden: This iteration is never released.  You should use the `use` function in Kotlin here:

```
override fun containsValue(value: FLValue): Boolean =
      dict.iterator().use { iter ->
          while (iter.key != null) {
              if (iter.value == value) return true
              iter.next()
          }
          false
      }
```

common/main/kotlin/com/couchbase/lite/internal/fleece/FLValueExtensions.kt
```diff
@@ -0,0 +1,107 @@
...
+            if (iter.value == value) return true
+            iter.next()
+        }
+        return false
+    }
+
+    class Entry(override val key: String, override val value: FLValue) : Map.Entry<String,FLValue>
+
+    override val entries: Set<Map.Entry<String, FLValue>> get() {
+        val iter = dict.iterator()
```
borrrden: This iterator is also never freed

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
borrrden: Even retaining body here is not going to work.  This needs to get translated into unmanaged memory.  The heart of the problem is that you are creating body as a Java object, which has the ability to be moved, collected, etc and then you have FLDict with a dangling pointer.  You can get an FLSliceResult object that you can then save by calling `finish2()` on the result of `encodeSerializableValue`

```
private fun <T:DocumentModel> MutableDocument.setContentFromModel(model: T, serializer: SerializationStrategy<T>) {
      val encoder = FLEncoder.getManagedEncoder()
      FleeceRootEncoder(encoder).encodeSerializableValue(serializer, model)
      val sliceResult = encoder.finish2()   // native malloc'd buffer — GC cannot move it
      setContent(sliceResult, false)        // Document stores sliceResult; buffer stays pinned
  }
```

common/main/kotlin/com/couchbase/lite/internal/fleece/FLValueExtensions.kt
```diff
@@ -0,0 +1,107 @@
...
+            if (array[i] == element) return true
+        return false
+    }
+
+    override fun containsAll(elements: Collection<FLValue>): Boolean =
+        elements.all { this.contains(it) }
+
+    override fun iterator(): Iterator<FLValue> {
+        return object: Iterator<FLValue> {
+            var iter = array.iterator()
```
borrrden: This iterator is never freed / closed.  It's going to leak.  The returned value may need to be turned into a `Closeable`.  So that it can be properly freed.

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
```diff
@@ -0,0 +1,176 @@
...
+        val curModel = curDoc?.let {modelFromC4Doc(this, it.id, it.c4doc, deserializer)}
+        val ok = conflictHandler!!(model, curModel)
+        if (ok)
+            doc?.setContentFromModel(model, serializer)
+        return ok
+    }
+
+    if (doc.revisionID != meta?.revisionID) {
+        // Model is out of date -- have to resolve the conflict
+        if (!handleConflict(null, doc))
```
borrrden, line 105: There is no null checking for conflictHandler here.  That worries me a bit.

common/main/kotlin/com/couchbase/lite/internal/fleece/FleeceDeserialization.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    override fun decodeBoolean(): Boolean   = value.asBool()
+    override fun decodeByte(): Byte         = decodeLong().toByte()
+    override fun decodeShort(): Short       = decodeLong().toShort()
+    override fun decodeChar(): Char         = Char(decodeInt())
+    override fun decodeInt(): Int           = decodeLong().toInt()
+    override fun decodeLong(): Long         = value.asInt()
+    override fun decodeFloat(): Float       = value.asFloat()
+    override fun decodeDouble(): Double     = value.asDouble()
+    override fun decodeString(): String     = value.asString()
```
borrrden: This needs to be nullable right?  If value is not a string type `asString()` I believe returns null.

borrrden (changes requested):

common/main/kotlin/com/couchbase/lite/internal/fleece/FleeceDeserialization.kt
snej: `decodeString()` is a Kotlin serialization method so I can't change its signature.

`value.asString()` is declared `@NonNull`, but the doc-comment says "Returns the exact contents of a string value, or null for all other types. ???" ¯\\_(ツ)_/¯

common/main/kotlin/com/couchbase/lite/CollectionExtensions.kt
snej: Thanks. But this passes the buck to the `Document.setContent` method: I don't know how to implement it to take an `FLSliceResult`. Where does it store that?

borrrden: I think it has to be added as another field right?

common/main/kotlin/com/couchbase/lite/internal/fleece/FleeceDeserialization.kt
borrrden: The implementation returns null so it looks to be incorrectly attributed.  I don't know what is appropriate here but there are two choices

1. Throw an exception
2. Coalesce to empty string

common/main/java/com/couchbase/lite/Document.java
```diff
@@ -622,6 +629,15 @@ private void setC4Document(@Nullable C4Document c4doc, boolean mutable) {
         }
     }
 
+    // for use by CollectionExtensions.kt
+    void setContent(@NonNull FLSliceResult fleeceData, boolean mutable) {
+        synchronized (lock) {
+            var data = FLValue.fromData(fleeceData).asFLDict();
+            extraBackingStore = fleeceData;
```
borrrden, line 636: This is a memory leak right?  This needs to be freed at some point (probably both in the close logic as well as here if the previous data is not null)

borrrden (changes requested):

common/main/java/com/couchbase/lite/Document.java
borrrden: Ok it's actually not a leak, but it's preferred to call the `close()` method on it rather than waiting for the finalizer

snej: I thought of that, but there's no place to call close() from, since Document doesn't have a close method.

borrrden: Hmmmm ok fair point, this is worth a comment on it then explaining that at some point we want to be able to call close on it.

borrrden (approved):
I'd like to see a comment on the place where we can't reach ideal behavior in explicitly freeing the FLSliceResult data on document.

PR #485, opened 2026-05-18
CBL-8160 : Restore the close database timeout waiting for active processes

pasin:
* The previous fix reduce the timeout waiting for the active processes to 10 seconds which is too low (e.g. WebSocket close timeout is 9 seconds). Restore the timeout back to 30 seconds.

* Restore the previous logic that will try to close the database anyway even though the active processes are not all finished.

* Noted that the original code before the fix has a bug that the dabase close only retries once after getting BUSY error but wait for total 30 seconds then timeout after.

### DIFF
--- a/.github/workflows/warn_release.yml
+++ b/.github/workflows/warn_release.yml
@@ -1,5 +1,6 @@
 on:
   pull_request:
+    types: [opened]
     branches: ['**/release/**']
 
 jobs:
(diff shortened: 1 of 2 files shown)


pasin:
Fixed a test blocker reported by QE.

jianminzhao (approved):

borrrden (approved):

PR #488, opened 2026-05-21
CBL-8318: Fix executor concurrency and inside executor tracking

pasin:
- Reverted the default CBLExecutor to a fixed-size pool by using POOL_SIZE as both the core and max pool size. With corePoolSize = 0 and an unbounded LinkedBlockingQueue, ThreadPoolExecutor queues additional work instead of growing the pool, making the concurrent executor behave effectively single-threaded.

- Fixed ConcurrentExecutor.isInsideExecutor() to work with overlapping tasks by keeping the insideExecutor flag inside thread local, and ensure SerialExecutor clears its current-thread state in a finally block.

- Added tests covering isInsideExecutor() for both serial and concurrent executors.

### DIFF
--- a/common/main/java/com/couchbase/lite/internal/exec/CBLExecutor.java
+++ b/common/main/java/com/couchbase/lite/internal/exec/CBLExecutor.java
@@ -130,7 +130,7 @@ public class CBLExecutor extends ThreadPoolExecutor {
                 }
             });
 
-        allowCoreThreadTimeOut(true);
+        allowCoreThreadTimeOut(false);
 
         this.name = name;
     }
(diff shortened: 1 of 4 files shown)


common/main/java/com/couchbase/lite/internal/exec/CBLExecutor.java
```diff
@@ -97,7 +97,7 @@ public Stats(
     private float m2;
 
     // A fixed sized pool backed by a very large queue.
-    public CBLExecutor(@NonNull String name) { this(name, 0, POOL_SIZE, new LinkedBlockingQueue<>()); }
+    public CBLExecutor(@NonNull String name) { this(name, POOL_SIZE, POOL_SIZE, new LinkedBlockingQueue<>()); }
```
borrrden, line 100: I really don't like this because it causes a lot of thread creation spamming.  Is there an issue that caused you to change

borrrden (approved):

common/main/java/com/couchbase/lite/internal/exec/CBLExecutor.java
pasin: As discuss, I have set `allowCoreThreadTimeOut` to mitigate the problem.

PR #494, opened 2026-06-05
Add back 3.x API for 3.4 release

borrrden:


borrrden:
This was the agreed upon way to handle 3.4 API

common/main/kotlin/com/couchbase/lite/CommonConfigurationFactories.kt
```diff
@@ -31,7 +32,30 @@ val CollectionConfigurationFactory: CollectionConfiguration? = null
...
+
+    (channels ?: this?.channels)?.let { config.channels = it }
+    (documentIDs ?: this?.documentIDs)?.let { config.documentIDs = it }
+    (pushFilter ?: this?.pushFilter)?.let { config.pushFilter = it }
+    (pullFilter ?: this?.pullFilter)?.let { config.pullFilter = it }
+    (conflictResolver ?: this?.conflictResolver)?.let { config.conflictResolver = it }
+
+    return config
+}
+
```
pasin, line 58: Seem like the restoration moved `@see com.couchbase.lite.CollectionConfiguration` away. Can you also add it here?

common/main/java/com/couchbase/lite/AbstractDatabase.java
```diff
@@ -620,6 +641,315 @@ public boolean equals(Object o) {
...
     }
 
+    //---------------------------------------------
+    // Deprecated API methods
+    //---------------------------------------------
+
+    /**
+     * The number of documents in the default collection.
+     *
+     * @return the number of documents in the database, 0 if database is closed.
```
pasin: This method is supposed to return zero in 3.x, but it seems like it has been implemented like this since 3.2.x. I think we can just remove `0 if database is closed` from the comment.

pasin (approved):
I have two small comments about the API doc comments. Can you check that before merging?

Also, I haven't seen the PR in EE repo (Maybe you haven't push it yet?).

common/main/kotlin/com/couchbase/lite/CommonConfigurationFactories.kt
borrrden: This is the way it was in release/3.3

PR #495, opened 2026-06-08
CBL-8402 : Fix data race on SerialExecutor.currentThread

pasin:
Make currentThread an AtomicLong so isInsideExecutor() reads a consistent, visible value across pool threads. Clear it with compareAndSet in the finally block so a task that has already claimed the thread for the next run is not cleared.


pasin:
Fix an issue reported by QE for 4.1.0.

borrrden (approved):

